PR #1, opened 2018-11-14
EID-1075

JHjava:
- Re-write acceptance tests in Ruby 
- Set up docker infrastructure for tests to run 
- Remove all old acceptance tests and scripts 
- Change start-up to not follow logs 
- Added acceptance tests to pre-commit with a horrible sleep command so that the tests will only run once all services have started. This is only until I've added a healthcheck so we detect when a service is running.
- Delete unused .pre-commit-config.yaml file 
- Bump versions of Ruby to 2.5.3


proxy-node-acceptance-tests/Gemfile.lock
```diff
@@ -0,0 +1,81 @@
+GEM
```
tomrosier: This probably can be gitignored, if we need it to be statically version we can just add the version to the Gemfile

proxy-node-acceptance-tests/.ruby-version
```diff
@@ -0,0 +1 @@
+2.4.2
```
tomrosier: Could we use 2.5.3?

pre-commit.sh
```diff
@@ -1,4 +1,17 @@
 #!/usr/bin/env bash
 set -e
 
-./docker-gradle clean test
+./docker-gradle clean -x test
+
+./startup.sh
+
+echo "Waiting 60 seconds for services."
+sleep 60
```
tomrosier: Is this a temporary work around?

proxy-node-acceptance-tests/.rubocop.yml
```diff
@@ -0,0 +1,9 @@
+Style/FrozenStringLiteralComment:
```
tomrosier: Is this autogenerated? or is it Config?

proxy-node-acceptance-tests/docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '3.4'
+
+services:
+  selenium-hub:
+    image: selenium/standalone-firefox
+
+  test-runner:
```
tomrosier: Could this be named `acceptance-tests`?

proxy-node-acceptance-tests/Gemfile.lock
vixus0: I think checking in the `Gemfile.lock` is fine, since we're not pinning our gem versions.

proxy-node-acceptance-tests/.ruby-version
vixus0: Yeah why not use the latest and greatest? We should bump the `pki` stuff while we're at it.

pre-commit.sh
vixus0: Temporary workaround until we implement healthchecks in the services that will tell us when they're up and working.

proxy-node-acceptance-tests/.rubocop.yml
JHjava: Config

proxy-node-acceptance-tests/docker-compose.yml
JHjava: Sure

proxy-node-acceptance-tests/.ruby-version
JHjava: The latest it is!

startup.sh
```diff
@@ -19,4 +32,7 @@ pushd "${PN_PROJECT_DIR}/pki"
     "${PKI_OUTPUT_DIR}"
 popd
 
-docker-compose up $@
+docker-compose up $@ -d
+
+wait_for "Metadata" localhost:5000/Metadata
```
Wynndow: This is actually waiting for the stub-connector service, rather than the metadata service (which is on 7000). Is that right? Might be worth renaming if so.

JHjava: It's the metadata for the gateway and translator services. Once the healthchecks are added to the other services then I can do the same here with them and get rid of the sleep command pre-commit.

Wynndow: Cool. If this check is just to ensure that the service is up and running, it might be worth changing to `wait_for "stub-connector" localhost:5000/Metadata` though.

vixus0: Yeah this is waiting for stub connector to come up so it should probably be renamed.

JHjava: I'll clean this up once the healtchecks are implemented

PR #2, opened 2018-11-15
EID-1062: Implement new validations for HUB response in Gateway and Translator

Wynndow:


proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/HubResponseAppRuleTests.java
```diff
@@ -153,13 +157,8 @@ private Response extractEidasResponse(Response hubResponse) throws Exception {
...
 
     private javax.ws.rs.core.Response postHubResponseToProxyNode(Response hubResponse) throws URISyntaxException {
-        String encodedResponse = Base64.encodeAsString(marshaller.transformToString(hubResponse));
-        Form postForm = new Form().param(SamlFormMessageType.SAML_RESPONSE, encodedResponse);
-
-        return proxyNodeAppRule
-                .target("/SAML2/SSO/Response/POST")
-                .request()
-                .post(Entity.form(postForm));
+        postEidasAuthnRequest(eidasAuthnRequest);
```
Wynndow: These tests now always test the happy path, where the gateway has seen an eIDAS authn request before it sees the corresponding hub response. It might be worth adding a test to check that we get the expected exception if this isn't the case.

Wynndow:
I was going to approve this, with a suggestion to add a tech debt ticket, or something in the backlog, about finding a solution to this issue of using the ConcurrentSkipListSet.

But I can't, as I opened the PR...

vixus0:
We could think about using some kind of (temporary) user session to remember the AuthnRequest ID to validate against, rather than storing a list of all the IDs seen thus far but agreed this could be in a later ticket.

Wynndow:
@vixus0 @CharlesIC Have added a test for the unhappy path if you could take a look.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/saml/validation/components/RequestIdWatcher.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public void observe(AuthnRequest authnRequest) {
+        seenAuthnRequestIds.add(authnRequest.getID());
+    }
+
+    public boolean haveSeenRequestFor(Response response) {
+        return seenAuthnRequestIds.contains(response.getInResponseTo());
+    }
+
+    public void validateSeenRequestFor(Response response) {
```
CharlesIC: Wouldn't we want to remove the request ID from the set once we've validated it's been seen? Or do we periodically clear the set? Maybe it's not an issue seen as this is only a temporary solution. Do we have a ticket in the backlog to find & implement a long-term solution?

vixus0: Yeah that might be a good plan actually. Would prevent replay attacks (somewhat).

CharlesIC: Talked to Chris. We'll fix this before going to integration so fine for now.

Wynndow: If we remove the ID, it'll mean that if the user needs to repost the hub response to the gateway for any reason, they won't be able to. Is that an issue at all?
Simon mentioned potentially using a Guice cache, but it doesn't solve the issue about losing all the id's on reboot, or scaling.

vixus0:
Wait are we not removing the `requestId` from the list after a successful validation?

PR #6, opened 2018-11-16
EID-1111 Add healthchecks for proxy-node services 

JHjava:
- Remove sleep command from pre-commit
- Wait for healthchecks to return 200 before completing start-up.sh
- Rebuild docker image when starting up service
- Clean up classes and make them consistent for each service


JHjava:
@simonwo Done!

PR #7, opened 2018-11-21
EID-1088: Implement new validations for country request in Gateway

CharlesIC:
Remove duplicated test resources
  - Added slashes to the resource filename, as this denotes a path which can be found in other packages, not just the local one

Add new validations of AuthnRequest in Gateway:
  - We now validate a bunch of attributes and sub-elements of the received country AuthnRequest
  - Lots of tests of the above
  - Updated example XML to be valid with new validations
  - Added unit tests to EidasAuthnValidatorTest
  - Add app rule tests for new validations.


JHjava:
Was pre-commit run? I'm seeing ""code":400,"message":"Error handling authn request."

CharlesIC:
Yes, I've run all the tests - it works locally _on my machine_ and also all the tests passed on Travis.
Can you retry, @JHjava ?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/saml/validation/EidasAuthnRequestValidator.java
```diff
@@ -46,6 +84,18 @@ public void validate(AuthnRequest request) {
         spTypeValidator.validate(spTypeElement);
         loaValidator.validate(request.getRequestedAuthnContext());
         requestedAttributesValidator.validate(requestedAttributesElement);
+        assertionConsumerServiceValidator.validate(request);
+        comparisonValidator.validate(request.getRequestedAuthnContext());
+
+        if (!duplicateAuthnRequestValidator.valid(request.getID())) {
```
Wynndow: the `duplicateAuthnRequestValidator` keep track of request ID's in an in memory map. If we have more than one instance running this validation becomes flakey. Also, if an instance is restarted we lose the current list. This map will just continues to grow and consume memory. Same issues that https://govukverify.atlassian.net/browse/EID-1117 is looking to resolve.

Also, why 5 minutes as the time limit for checking duplication? Shouldn't the ID's be unique - why have a time limit?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/base/ProxyNodeAppRuleTestBase.java
```diff
@@ -94,21 +96,46 @@
...
+            fail(e);
+            return null;
+        }
+    }
+
+    protected Response redirectEidasAuthnRequest(AuthnRequest eidasAuthnRequest) {
+        String encodedRequest = Base64.encodeAsString(marshaller.transformToString(eidasAuthnRequest));
+
+        try {
+            return proxyNodeAppRule.target("/SAML2/SSO/Redirect")
```
Wynndow: What is the redirect endpoint used for?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/saml/validation/components/ComparisonValidatorTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private static RequestedAuthnContext aRequestedAuthnContext(AuthnContextComparisonTypeEnumeration comparison) {
+        RequestedAuthnContext requestedAuthnContext = new RequestedAuthnContextBuilder().buildObject();
+        requestedAuthnContext.setComparison(comparison);
+        return requestedAuthnContext;
+    }
+
+    @Test
+    public void shouldThrowIfAuthnContextIsMissing() {
+        assertThrows(InvalidAuthnRequestException.class, () -> validator.validate(null));
```
Wynndow: For some reason running the tests with gradle isn't picking this up, but with IntelliJ it is. `ComparisonValidator ` throws a `InvalidHubResponseException`, not a `InvalidAuthnRequestException ` when `authnContext` is missing I think?

Wynndow: Same thing appears in Travis' logs for the build

proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
```diff
@@ -252,15 +257,23 @@ private ResponseAssertionDecrypter createDecrypter(KeyPairConfiguration configur
...
-            new SpTypeValidator(),
-            new LoaValidator(),
-            new RequestedAttributesValidator()
+                new RequestIssuerValidator(),
+                new SpTypeValidator(),
+                new LoaValidator(),
+                new RequestedAttributesValidator(),
+                new DuplicateAuthnRequestValidator(new ConcurrentHashMap<>(), Duration.standardMinutes(5)),
+                new ComparisonValidator(),
+                createDestinationValidator(configuration),
```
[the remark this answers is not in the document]
simonwo: Just to abstract away how the configuration is used.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/saml/validation/EidasAuthnRequestValidator.java
simonwo: Yes this is the same problem. We decided to "fix it later" rather than sort it now. Sorting this is in scope for EID-1117.

AuthnRequests come from the service – I think the duplication window is there to allow the user's browser to resubmit if there are network errors, they press Back, etc. 5 minutes is a fairly aribtrary value that we need to pick a more intelligent value for in production.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/base/ProxyNodeAppRuleTestBase.java
simonwo: I think it's just another way of accessing a SAML journey. It allows you to use a GET rather than a POST? We don't support it for Verify but we have to for eIDAS.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/saml/validation/components/ComparisonValidatorTest.java
simonwo: Wow, geez. We should fix that.

simonwo: Turns out tests in `ComparisonValidatorTest` weren't being run because they used the wrong annotation. This has been fixed.

Wynndow: Interesting that the tests were run by IntelliJ - any idea why?

PR #17, opened 2018-12-11
EID-1155: Dockerize softHSM in Proxy Node Translator

dbes-gds:
Install softHSM in the Translator container so that the proxy node signing key can be stored in and retrieved from softHSM in test environments.
Modify startup and pki generation scripts to generate the signing key files in a format that can be loaded into softHSM.
Load the signing private key into softHSM in the Dockerfile after installing softHSM so that it is available for use during testing.
Include swedenconnect/opensaml-pkcs11 in the Translator build script so that the library can be used to access keys in softHSM.  Integration tests provided to prove that softHSM and swedenconnect/opensaml-pkcs11 are installed and usable in the container (built from translator.softHSM.Dockerfile).
The docker-compose files have been modified to use the new translator.softHSM.Dockerfile, and include additional arguments and environment variables used to load the key(s) into softHSM.
The pre-commit.sh script for Proxy Node uses a separate docker-compose file and image to compile and test the code.  This container does not use the individual proxy node component containers to run the tests, meaning that the softHSM installation was missing for pre-commit.  As such a new Dockerfile has been created to build this 'docker-gradle' service (pre-commit.softHSM.Dockerfile) which includes the softHSM installation too.  This means that the softHSM integration tests are still able to run under pre-commit.

In summary, these changes install and test softHSM and swedenconnect/opensaml-pkcs11 in the Translator Docker container, and prove the ability to load our test keys into softHSM.

The use of a key retrieved from softHSM to sign messages in the proxy node is to be implemented in a subsequent task EID-1160.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,7 @@ ext {
     dropwizard_version = '1.2.0'
     utils_version = '2.0.0-346'
     saml_libs_version = "${opensaml_version}-160"
+    soft_hsm_version = '1.1.1'
     build_version = "$opensaml_version-${System.env.BUILD_NUMBER ?: 'SNAPSHOT'}"
 }
 
@@ -47,6 +48,7 @@ subprojects {
         dropwizard
         saml
         saml_test
+        soft_hsm
     }
 
     dependencies {
@@ -80,6 +82,10 @@ subprojects {
                 "javax.activation:activation:1.1.1",
                 "io.lettuce:lettuce-core:5.1.3.RELEASE",
                 "uk.gov.ida:security-utils:${utils_version}"
+        )
+
+        soft_hsm (
+                "se.swedenconnect.opensaml:opensaml-pkcs11-support:${soft_hsm_version}"
         )
 
         testCompile (
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -28,10 +28,19 @@ services:
     volumes:
       - ./.local_pki:/verify-eidas-proxy-node/pki:ro
   proxy-node-translator:
+    depends_on:
+      - soft-hsm
     build:
       context: .
+      dockerfile: translator.softHSM.Dockerfile
       args:
         component: proxy-node-translator
+        softHSMSigningKeyLabel: vfpn-uk-signing
+        softHSMSigningKeyToken: vfpn-uk-signing
+        softHSMSigningKeySlot: 0
+        softHSMSigningKeyId: DEF0
+        softHSMSigningKeyPin: CHANGE-ME-123
+        softHSMSigningKeySoPin: CHANGE-ME-456
     ports:
       - "6300:80"
       - "6661:6661"
@@ -47,9 +56,17 @@ services:
       CONNECTOR_NODE_METADATA_URL: http://stub-connector/Metadata
       CONNECTOR_NODE_ENTITY_ID: http://localhost:5000/Metadata
       REDIS_SERVER_URI: redis://gateway-redis:6379/
+      SOFT_HSM_LIB_PATH: /usr/lib/softhsm/libsofthsm2.so
     env_file: ./.local_pki/proxy_node.env
     volumes:
-    - ./pki:/verify-eidas-proxy-node/pki:ro
+      - ./pki:/verify-eidas-proxy-node/pki:ro
+  soft-hsm:
+    build:
+      context: .
+      dockerfile: softHSM.baseImage.Dockerfile
+      args:
+        baseImage: openjdk:11-jre-slim
+    image: govukverify/soft-hsm
   stub-connector:
     build:
       context: .
--- a/gradle-compose.yaml
+++ b/gradle-compose.yaml
@@ -2,12 +2,17 @@ version: '3'
 
 services:
   docker-gradle:
-    image: gradle:jdk11
+    build:
+      context: .
+      dockerfile: softHSM.baseImage.Dockerfile
+      args:
+        baseImage: gradle:jdk11
     working_dir: /home/gradle/project
     user: root
     entrypoint: gradle --no-daemon
     environment:
       VERIFY_USE_PUBLIC_BINARIES: 'true'
+      SOFT_HSM_LIB_PATH: /usr/lib/softhsm/libsofthsm2.so
     volumes:
       - gradle-cache:/home/gradle/.gradle
       - .:/home/gradle/project
--- a/pki/options.rb
+++ b/pki/options.rb
@@ -2,7 +2,20 @@ require 'optparse'
 
 USAGE = 'Usage: generate.rb [options] <output directory>'
 
-Options = Struct.new(:hub_entity_id, :idp_entity_id, :proxy_entity_id, :hub_response_url, :idp_sso_url, :proxy_sso_url, :do_files, :do_manifests, :do_env, :xmlsectool_path, :truststore_pass)
+Options =
+    Struct.new(
+        :hub_entity_id,
+        :idp_entity_id,
+        :proxy_entity_id,
+        :hub_response_url,
+        :idp_sso_url,
+        :proxy_sso_url,
+        :do_files,
+        :do_manifests,
+        :do_env,
+        :xmlsectool_path,
+        :truststore_pass,
+        :softhsm)
 
 class Parser
   def self.parse(args)
@@ -33,6 +46,7 @@ class Parser
       opts.on('--env', "Output environment files for Docker Compose") { |_| options.do_env = true }
       opts.on('--xmlsectool PATH', "Path to xmlsectool (default: xmlsectool)") { |s| options.xmlsectool_path = s }
       opts.on('--truststore-pass PASSWORD', "Password for generated truststores (default: marshmallow)") { |s| options.truststore_pass = s }
+      opts.on('--softhsm', "Use softhsm for testing") { |_| options.softhsm = true }
       opts.on('-h', '--help', 'Print help message') { |_| abort(opts.to_s) }
     end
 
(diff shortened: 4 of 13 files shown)


translator.softHSM.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+ENTRYPOINT ["gradle", "--no-daemon"]
+
+# Runtime Image
+FROM openjdk:11-jre-slim
+
+USER root
+ENV OPEN_SC_VERSION=0.19.0
+
+# Install softHSM and OpenSC
+RUN apt-get update && apt-get install -y softhsm pcscd libccid libpcsclite-dev libssl-dev libreadline-dev autoconf automake build-essential docbook-xsl xsltproc libtool pkg-config wget && \
```
tomrosier: Having all the commands compacted together will potentially make docker slower as it has todo all theses tasks every time the image is rebuilt.

translator.softHSM.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+USER root
+
+ARG component
+COPY --from=build /app/${component}/build/install/${component} /app
+
+# Copy test signing keys
+RUN mkdir /app/keys
+COPY --from=build /app/${component}/src/dist/proxy_node_signing.* /app/keys/
+
+# softHSM Environment Variable assignment from ARGS
```
tomrosier: This stuff should be at the top really, as its unlikely to change. Which will speed up builds

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Tue Dec 11 10:37:58 GMT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-all.zip
```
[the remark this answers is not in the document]
dbes-gds: The 'all' version of gradle includes extra debugging tools for the ide.  Have reset this to 'bin'.

pre-commit.softHSM.Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Docker Image for pre-commit.sh to run all tests.
+# softHSM tests require softHSM and OpenSC.
+#
+FROM gradle:jdk11
+USER root
+
+ENV OPEN_SC_VERSION=0.19.0
+
+# Install softHSM and OpenSC
+RUN apt-get update && apt-get install -y softhsm pcscd libccid libpcsclite-dev libssl-dev libreadline-dev autoconf automake build-essential docbook-xsl xsltproc libtool pkg-config wget && \
```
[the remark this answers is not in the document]
dbes-gds: Have split into two separate RUNS, one for libs and one for OpenSC.  Don't think we can gain much more as this image will be fairly static and won't be rebuilt even if the code changes.

translator.softHSM.Dockerfile
dbes-gds: Have created a separate docker image for the softhsm installation and included it as the runtime image so that this installation does not have to happen every time there is a build.

translator.softHSM.Dockerfile
dbes-gds: Improve by use of the softhsm-only image.

pki/options.rb
```diff
@@ -2,7 +2,7 @@
 
 USAGE = 'Usage: generate.rb [options] <output directory>'
 
-Options = Struct.new(:hub_entity_id, :idp_entity_id, :proxy_entity_id, :hub_response_url, :idp_sso_url, :proxy_sso_url, :do_files, :do_manifests, :do_env, :xmlsectool_path, :truststore_pass)
+Options = Struct.new(:hub_entity_id, :idp_entity_id, :proxy_entity_id, :hub_response_url, :idp_sso_url, :proxy_sso_url, :do_files, :do_manifests, :do_env, :xmlsectool_path, :truststore_pass, :softhsm)
```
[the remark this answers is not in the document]
dbes-gds: Apparently not :)

translator.softHSM.Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+COPY build.gradle build.gradle
+COPY settings.gradle settings.gradle
+COPY proxy-node-shared/ proxy-node-shared/
+
+ARG component=proxy-node-gateway
+COPY ${component}/src ${component}/src
+COPY ${component}/build.gradle ${component}/build.gradle
+
+RUN gradle --no-daemon --quiet -p ${component} --parallel install -x test
+ENTRYPOINT ["gradle", "--no-daemon"]
```
Wynndow: From this line up, and from line 58 down, the Dockerfiles are the same as the base Dockerfile used to build the other images. Instead of the duplication, could you extend that Dockerfile with the extra bits here?

softHSM.runtime.Dockerfile
```diff
@@ -0,0 +1,21 @@
+#
```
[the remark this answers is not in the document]
dbes-gds: The pre-commit one needs gradle to build whereas the runtime one doesn't, means runtime can be slimmer.

proxy-node-translator/src/test/java/uk/gov/ida/notification/softhsm/SoftHSMIntegrationTest.java
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * softHSM tests should only run if softHSM is installed and configured.
+     * Installation and configuration to be indicated by the SOFT_HSM_LIB_PATH environment variable.
+     * softHSM is installed in the Translator Docker image and this variable is set, so the tests
+     * will run on pre-commit.  Running the tests when building locally is effectively excluded
+     * by not setting this environment variable.
+     */
+    @Before
+    public void checkSoftHSMConfigured() {
+        org.junit.Assume.assumeTrue(softHSMLibPath != null && softHSMLibPath.length() > 0);
```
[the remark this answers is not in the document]
dbes-gds: fixed

pre-commit.softHSM.Dockerfile
```diff
@@ -0,0 +1,21 @@
+#
+# Docker Image for pre-commit.sh to run all tests.
+# softHSM tests require softHSM and OpenSC.
+# Gradle required for code build.
+#
+FROM gradle:jdk11
```
Wynndow: This Dockerfile is the same as the runtime one, apart from this line. Is there any way to reduce the replication?

build.gradle
```diff
@@ -82,6 +84,10 @@ subprojects {
                 "uk.gov.ida:security-utils:${utils_version}"
         )
 
+        soft_hsm (
```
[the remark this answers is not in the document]
dbes-gds: tried it and it didn't work, the reference to the configuration is in the translator gradle, but the configuration definition had to go in the top level.  It kind of makes sense as that is where all the versions are defined.

proxy-node-translator/src/test/java/uk/gov/ida/notification/softhsm/SoftHSMIntegrationTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+
+    @Test
+    public void shouldRetrieveCredentialsFromHSM() throws Exception {
+
+        PKCS11ProviderConfiguration configuration = createPKCS11SoftHSMProviderConfigurationWithCredentials(CRED_ALIAS, CRED_ENV, CRED_PIN, "162624334", CRED_NAME);
+        PKCS11Provider pkcs11Provider = getPkcs11ProviderForConfiguration(configuration);
+        TestKeyPair signingTestKeyPair = new TestKeyPair(PROXY_NODE_SIGNING_CRT, PROXY_NODE_SIGNING_PK_8);
+
+        addPrivateKeyToKeyStore(signingTestKeyPair, CRED_ALIAS, CRED_PIN);
```
Wynndow: What's happening here? `addPrivateKeyToKeyStore ` doesn't seem to return anything - does it  mutate the `signingTestKeyPair` in some way?

pre-commit.softHSM.Dockerfile
Wynndow: I think you could reduce the two files to one. You could set an `ARG` before the `FROM` that you set to the image you want to base them off (either `openjdk:11-jre-slim`, or  `gradle:jdk11`), and then use that arg in the `FROM` directive. You just need to set the arg to the right thing in the docker-compose and gradle-compose files.

gradle-compose.yaml
```diff
@@ -2,12 +2,17 @@ version: '3'
...
+      context: .
+      dockerfile: softHSM.baseImage.Dockerfile
+      args:
+        baseImage: gradle:jdk11
     working_dir: /home/gradle/project
     user: root
     entrypoint: gradle --no-daemon
     environment:
       VERIFY_USE_PUBLIC_BINARIES: 'true'
+      SOFT_HSM_LIB_PATH: /usr/lib/softhsm/libsofthsm2.so
```
tomrosier: Is this likely to change?

docker-compose.yaml
```diff
@@ -47,9 +56,17 @@ services:
       CONNECTOR_NODE_METADATA_URL: http://stub-connector/Metadata
       CONNECTOR_NODE_ENTITY_ID: http://localhost:5000/Metadata
       REDIS_SERVER_URI: redis://gateway-redis:6379/
+      SOFT_HSM_LIB_PATH: /usr/lib/softhsm/libsofthsm2.so
```
tomrosier: Is this likely to change?, Can we put it in the docker file if it will stay the same 🙂

PR #21, opened 2018-12-14
EID-1150 - Local minikube deployment

chrisfarms:
## What

* Adds "Charts" for basic deployments of gateway/translator/stubs to kubernetes
* Updates pki generation script to output Kubernetes ConfigMaps suitable for referencing from pods
* Updates startup.sh script to provision a single-node kubernetes cluster (minikube), build the application container images and deploy the charts
* Tweaks acceptance test script so that it targets the correct

## Known issues

The following are know issues that we'll be addressed in a follow up story about improving the Charts:

* Race condition between pki configmap vs pods starting - this is caused by a lack of liveness/readiness probes - workaround: `kubectl delete` each of the running pods to kick them
* PKI as configmap is not ideal - we intend to improve this situation
* Race condition between ./startup.sh and acceptance tests running - will be solved once readiness probes exist making it possible to wait for the system to be ready


startup.sh
```diff
@@ -2,40 +2,67 @@
...
+# fi
+
+# build all the images locally
+# send all the images to the dockerdaemon in minikube
+mkdir -p "${BUILD_DIR}"
+for component in $COMPONENTS; do
+	tag="local-$(tar c $component | md5sum | awk '{print $1}')"
+	image="govukverify/${component}:${tag}"
+	echo "building $component as $image"
+	if (eval $(minikube docker-env) && docker inspect --type=image "${image}" >/dev/null 2>&1); then
```
vixus0: `minikube docker-env` uses the `SHELL` environment variable to generate its output. This breaks for anyone not using `bash` as their shell because it seems to inherit the `SHELL` variable from the environment. So if I'm using `fish`, `minikube docker-env` seems to produce output using fish syntax even though the script is running in `bash`. So might want to stick an `export SHELL=/bin/bash` at the top of the script just to be safe.

chrisfarms: Resolved by explicitly asking for the bash script from `minikube docker-env --shell=bash`

startup.sh
```diff
@@ -2,40 +2,67 @@
...
-
-wait_for "Gateway"  localhost:6601/healthcheck
-wait_for "Translator" localhost:6661/healthcheck
-wait_for "Stub Connector" localhost:6667/healthcheck
+PKI_OUTPUT_DIR="${PN_PROJECT_DIR}/${PKI_DIR}"
+
+# boot a local k8s cluster
+(minikube status | grep -i running) || minikube start --memory 4096
+
+# clean
```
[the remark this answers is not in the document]
chrisfarms: stripped out comments

PR #23, opened 2018-12-17
EID-1163: Add more timeout options to translator client

vixus0:
The `connectionTimeout` alone wasn't enough to deal with the slowness
of speaking to the translator when it first starts up.

### DIFF
--- a/proxy-node-gateway/src/dist/config.yml
+++ b/proxy-node-gateway/src/dist/config.yml
@@ -31,7 +31,9 @@ replayChecker:
 translatorService:
   url: ${TRANSLATOR_URL}
   client:
+    timeout: ${TRANSLATOR_CONNECTION_TIMEOUT:-5s}
     connectionTimeout: ${TRANSLATOR_CONNECTION_TIMEOUT:-5s}
+    connectionRequestTimeout: ${TRANSLATOR_CONNECTION_TIMEOUT:-5s}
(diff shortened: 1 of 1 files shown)


proxy-node-gateway/src/dist/config.yml
```diff
@@ -31,7 +31,9 @@ replayChecker:
 translatorService:
   url: ${TRANSLATOR_URL}
   client:
+    timeout: ${TRANSLATOR_CONNECTION_TIMEOUT:-5s}
     connectionTimeout: ${TRANSLATOR_CONNECTION_TIMEOUT:-5s}
+    connectionRequestTimeout: ${TRANSLATOR_CONNECTION_TIMEOUT:-5s}
```
Wynndow: Where does this get used? I've grepped the proxy-node repo and can only find it mentioned here in the config?

vixus0: The `client:` block gets deserialised into a `JerseyClientConfiguration` which gets provided to the `TranslatorService` that gets built by `TranslatorServiceConfiguration`. The `*timeout:` options are standard params for a Jersey client as per [this Dropwizard class](/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientConfiguration.java).

PR #24, opened 2018-12-20
EID-1184: Build softhsm + pkcs11-proxy image

vixus0:
Creates a softhsm + pkcs11-proxy image using `softhsm/Dockerfile`.
Running a `softhsm` container will expose the `softhsm`'s PKCS11 API over
TCP port 5656. The API can then be consumed elsewhere by using the
`libspkcs11-proxy.so` module under `/softhsm` and setting the environment
variable `PKCS11_PROXY_SOCKET`. See `softhsm/docker-compose.yml` for
examples.

At the moment I'm just generating an ECDSA private key and certificate
and then importing the private key into the HSM when you build the
softhsm container. The key gets deleted by the cert is exposed for use
in our apps.

I've also symlinked `<component>/Dockerfile` to the root `Dockerfile`
for our Java apps. This means `startup.sh` can build everything
including the `softhsm` image.


softhsm/Dockerfile.tester
```diff
@@ -0,0 +1,3 @@
+from alpine
```
tomrosier: Is this dockerfile used?

vixus0: Yeah, by the docker-compose if you'd like to test.

PR #26, opened 2019-01-04
EID-1204: Local startup and readiness probes

chrisfarms:
## What

* Adds liveness probes to deployment charts so that pods get restarted if their /healthcheck endpoints fail
* Adds readiness probes to deployment charts so that no traffic is routed to pods that are failing healthcheck
* Adds logic to local `startup.sh` to wait for all liveness/readiness probes to report healthy before exit
* Tidies up some of the previous softhsm work as new builds of translator were failing to startup and there was a strange dependancy between docker-gradle and the translator's base image.

## Why

* Removes the race condition between `./startup.sh` and acceptance test execution
* Allows kubernetes to heal sick pods

## How to review

* Code review
* Ensure ./pre-commit.sh is working as expected

If you want to nuke everything to get a clean slate to test against:

```
minikube delete # destroy your local k8s vm
rm -rf .local_yaml # destroy the last build local deployment definition
rm -rf .local_pki # destroy the last built local pki
docker rmi -f $(docker images -qa) # removes ALL docker images and anything running 
```


charts/proxy-node-gateway/templates/gateway-deployment.yaml
```diff
@@ -25,8 +25,22 @@ spec:
...
+          containerPort: 80
+        - name: mgmt
+          containerPort: 6601
+        livenessProbe:
+          httpGet:
+            path: /healthcheck
+            port: mgmt
+          initialDelaySeconds: 20
+          periodSeconds: 5
+        readinessProbe:
```
paroxp: We should probably bring this back to the team and discuss potential replacement as the readiness and liveness probes differ.

chrisfarms: I was gonna leave this for if graceful shutdown was ever implimented for the apps but after looking into how the apps handle SIGINT it seems they stop accepting connections and wait for some threads to die ... so a simple http readiness check will do the trick ... updated

idrop:
I've successfully run the 'federation' on this branch, by also following local env setup at https://hackmd.io/sUek7RSdQMaqTCxr3fxLHw?view, but as a new starter with little context ATM, I don't feel confident to approve. Convince me otherwise?!

paroxp:
> Convince me otherwise?!

Challenge accepted. - This is relatively new to eIDAS team, meaning, you're now the person with most of the context. :D

/cc @idrop

PR #29, opened 2019-01-15
EID-1184: Move softhsm into a separate container

vixus0:
This makes `softhsm` a separate container in the `proxy-node-translator` deployment. Probably easiest to review this PR commit-by-commit.


startup.sh
```diff
@@ -4,48 +4,67 @@ set -euo pipefail
...
 for component in $COMPONENTS; do
 	tag="local-$(tar c $component | md5sum | awk '{print $1}')"
 	image="govukverify/${component}:${tag}"
 	echo "building $component as $image"
 	if (eval $(minikube docker-env --shell bash) && docker inspect --type=image "${image}" >/dev/null 2>&1); then
 		echo "already built"
 	else
-		docker build --file "${component}/Dockerfile" --build-arg "component=${component}" -t "${image}" .
-		docker save "${image}" | (eval $(minikube docker-env --shell bash) && docker load)
+  docker build --file "$component/Dockerfile" --build-arg "component=${component}" -t "${image}" .
```
tomrosier: Indentation has gone funny on this file.

softhsm/Dockerfile
```diff
@@ -1,27 +1,33 @@
...
 run wget https://github.com/SUNET/pkcs11-proxy/archive/master.zip && \
     unzip master.zip && \
     cd pkcs11-proxy-master && \
     cmake . && make && make install
 
 # Run image
-from alpine:latest
-workdir /softhsm
+from openjdk:11-jre-slim
+workdir /work
```
tomrosier: We should probably have a more descriptive name for the application director

softhsm/init.sh
```diff
@@ -0,0 +1,28 @@
...
+pkcs11-tool --module $soft_hsm_lib -p $SOFT_HSM_SIGNING_KEY_PIN -l -w signing_cert.crt -y cert -a $SOFT_HSM_SIGNING_KEY_LABEL -d $SOFT_HSM_SIGNING_KEY_ID
+
+# Clean up PKI
+rm signing_private_key.p8 signing_cert.crt
+
+# Copy the required library file
+cp /usr/local/lib/libpkcs11-proxy.so /softhsm/
+
+# Run pkcs11-proxy on port 5656
+exec "/usr/local/bin/pkcs11-daemon" "$soft_hsm_lib" $@
```
tomrosier: Is there any chance of SeLinux or Ip-tables causing issues with this port?

startup.sh
vixus0: That's because someone had converted everything to tabs >:o

softhsm/init.sh
vixus0: Not within a container, I should think. Been running fine on k8s.

softhsm/Dockerfile
vixus0: I'm going to stick with the `/app` convention we have on the Java `Dockerfile`s, if that's ok.

proxy-node-translator/Dockerfile
```diff
@@ -12,41 +9,20 @@ COPY build.gradle build.gradle
...
-ARG component=proxy-node-translator
+ARG component
 COPY --from=build /app/${component}/build/install/${component} /app
-COPY --from=build /app/${component}/src/dist/init.sh /bin/init.sh
 
 ENV CONFIG_FILE config.yml
 ENV COMPONENT $component
-ENTRYPOINT ["/bin/init.sh"]
-CMD []
+CMD "bin/$COMPONENT"
```
chrisfarms: Is there a reason why you want to pass `bin/$COMPONENT` as a command to whatever the base image's `ENTRYPOINT` is rather than running the process directly?

This can have impacts on how signals are propagated to the process and can make it harder to reason about the process tree within the container. I found it's good practice to set the desired `ENTRYPOINT` to and clear out any `CMD`from the parent image to make it very explict what you are running.

softhsm/Dockerfile
```diff
@@ -1,27 +1,33 @@
...
+workdir /app
 
 # Build softHSM source
 arg version=2.5.0
-run apk --update add opensc build-base wget sqlite-dev openssl-dev
+
+run apt-get update && \
+    apt-get install -y \
+    opensc build-essential wget libsqlite3-dev libssl-dev \
+    cmake libseccomp-dev libengine-gost-openssl1.1
```
tomrosier: `libengine-gost-openssl1.1` seems to be missing from Debian Stretch, this can be fixed by adding `--disable-gost` to `./configure`

proxy-node-translator/Dockerfile
vixus0: Yeah, I realised signals weren't being handled properly so I'm trying `dumb-init` for the `ENTRYPOINT` and leaving the `CMD` as-is. The main reason was so one could debug the container using `docker run` without having to override the entrypoint.

PR #30, opened 2019-01-21
Replace Minikube and use the docker-machine Kubernetes

tomrosier:
This replaces Minikube with the built in Kubernetes that is included with the desktop version of Docker.

To switch use this config.
```
kubectl config use-context docker-for-desktop
docker-machine create default
docker-machine env default
```


tomrosier:
Don't think this is relevant anymore

PR #31, opened 2019-01-21
EID-1228: Implement a generic error page.

tomrosier:
Implement a simple error page,

It shows a error page now when there is an error with the message returned from the hub.

Added extra tests to.

Screenshot of the error page. (its just the stub IDP one for now)

screen shot 2019-01-22 at 16 16 03


Wynndow:
Is this a user facing thing? Can you stick a screenshot of the error page up?

tomrosier:
> Is this a user facing thing? Can you stick a screenshot of the error page up?

Updated the description.

Wynndow:
If this is potentially going to be in front of users, does it need content/design?

proxy-node-gateway/src/main/resources/uk/gov/ida/notification/views/error-page.mustache
```diff
@@ -0,0 +1,50 @@
...
+        margin-top: 20px;
+        margin-bottom: 20px;
+    }
+    .help {
+        margin-top: 20px;
+        margin-bottom: 20px;
+    }
+</style>
+<body class="main">
+<div class="logo">{{reaction}}</div>
```
Wynndow: How do the variables get passed to the template? I can see where they're coming from in the ErrorPageView, but is there some magic that passes them to moustache?

tomrosier: Moustache does some magic behind the scenes, it uses the getters on the `ErrorPageView` so for a variable called `reaction` it uses the getter `getReaction`.

tomrosier:
> If this is potentially going to be in front of users, does it need content/design?



> If this is potentially going to be in front of users, does it need content/design?

Its a discussion that needs to be had, It probably be better if we can use the error pages hosted by the frontend rather than have the gateway serve the error page.

CharlesIC:
As we discussed with @tomrosier and @dbes-gds , we don't actually need this as the IDP errors are handled in the frontend/hub and won't get forwarded to the Proxy Node

JHjava:
@Wynndow This change was originally just meant to handle the error request, rather than create a page. There are additional tickets for that, if it is needed. @CharlesIC Good spot, what about errors generated in the Hub rather than the IDP? Let's discuss in person today.

CharlesIC:
@Wynndow I believe any error generated in the Hub will be handled there, the user should always end up on an error page if something goes wrong and I think we only send a response to the RP (which what PN effectively is) if everything is ok. We only need to validate the auth response from the hub - so the usual SAML validation, etc and potentially an error page there

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/HubResponseAppRuleTests.java
```diff
@@ -48,6 +48,7 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.opensaml.saml.saml2.core.StatusCode.*;
```
vixus0: Since it's in a test it's probably not such a big deal but we should avoid `*` imports.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
```diff
@@ -144,8 +141,7 @@ private void registerProviders(Environment environment) {
     }
 
     private void registerExceptionMappers(Environment environment) {
-        environment.jersey().register(new HubResponseExceptionMapper());
-        environment.jersey().register(new AuthnRequestExceptionMapper());
+        environment.jersey().register(new GenericExceptionMapper());
```
idrop: The two exception mappers the GenericExceptionMapper replaces both log the exception, while GenericExceptionMapper does not. Is this intentional?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/mappers/GenericExceptionMapper.java
```diff
@@ -0,0 +1,15 @@
...
+
+import uk.gov.ida.notification.views.ErrorPageView;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class GenericExceptionMapper implements ExceptionMapper<Exception> {
+    @Override
+    public Response toResponse(Exception exception) {
+        return Response.status(Response.Status.BAD_REQUEST)
```
idrop: Should we be setting status 400 in the case of a legitimate 500 error?

idrop:
Looks good, 3 response mappers with GenericeExceptionMapper as the fall-through exception mapper

PR #38, opened 2019-01-28
Bau - fix acceptance tests

JHjava:
- Call the run-tests-with-docker script when running pre-commit. The pre-commit script in verify-acceptance-tests that it was calling before didn't work due to variables not being set. Now we can run tests in docker we should be using that anyway and there is no need for both.
- Upgrade gems


JHjava:
@idrop That's an interesting point. I would have thought it would be better if it was left up running. Reason being, is if the tests fail you may want to make a small change to fix it and then run ./start-up.sh to build that component rather than bringing everything down and up again.

PR #50, opened 2019-02-04
EID-1278 - Output logs in Json format

JHjava:
- Bring in logstash depdency to prevent having to bring in ida-utils
- Output logs to StdOut in JSON so they can be parsed correctly by fluentd 
- Don't output to files as they will be lost when the containers are destroyed.

 **How to review**
- Run pre-commit 
- Go into the Stub Connector, Gateway and Translator pods  - kubectl exec -it '_pod-name_' bash
- Check the logs are in json format by 'kubectl logs '_pod-name_''

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,8 +17,7 @@ apply plugin: 'com.github.ben-manes.versions'
 ext {
     opensaml_version = '3.4.0'
     dropwizard_version = '1.2.0'
-    test_utils_version = '2.0.0-346'
-    utils_version = '2.0.0-336'
+    utils_version = '2.0.0-346'
     saml_libs_version = "${opensaml_version}-163"
     soft_hsm_version = '1.1.1'
     build_version = "$opensaml_version-${System.env.BUILD_NUMBER ?: 'SNAPSHOT'}"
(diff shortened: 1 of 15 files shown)


willp-bl:
if you want json logs without pulling in ida-utils you could look at using https://github.com/alphagov/dropwizard-logstash/blob/master/src/main/java/uk/gov/ida/dropwizard/logstash/LogstashBundle.java directly.  Is there a logging bundle in ida-utils or is an old version of dropwizard-logstash being pulled in transitively?

JHjava:
@willp-bl Good suggestion

eidas-saml-parser/src/dist/config.yml
```diff
@@ -8,10 +8,16 @@ server:
 
 logging:
   level: INFO
-  loggers:
-    uk.gov.ida: DEBUG
   appenders:
     - type: console
+      layout:
+        type: json
```
willp-bl: any reason to use a different lib for these logs? could use logstash-console from dropwizard-logstash to match the other logging outputs

dropwizard-logstash is set up so the logs have standardised keys in our logging saas

JHjava: Wanted to output to StdOut in JSON format rather than multilined. I can still use logstash-console and that should keep the standardised keys. logstash-file will output the logs to json in a file but unable to configured StdOut to do so without the dropwizard-json lib.

build.gradle
```diff
@@ -95,6 +99,8 @@ subprojects {
 
         verify_saml_test (
                 "uk.gov.ida:common-test-utils:${utils_version}",
+                "net.sourceforge.htmlunit:htmlunit:2.28",
```
vixus0: This doesn't look right. Why do we need to pull in `htmlunit`? Also why is `common-test-utils` being pulled in with two version numbers?

JHjava: Yep, that isn't right.

JHjava: This happened when I rebased

PR #55, opened 2019-02-08
EID-1304 - Set up the ESP the eidas authn request validator

JHjava:
- The ESP will be now be responsible for validating the eIDAS authn request
from the connector code. Set up the ESP with the classes and config needed
to carry out this validation but don't tie it into the resource yet. Will do that in a separate PR. 
- Add tests for the ESP validation
- Put in the bones in the Application class. 
- Place variables in config.yml as a placeholder, these aren't currently being set anywhere. The DefaultNotUsed will be removed once they are. 
- Add connector metadata configuration in the ESP, so the connector metadata bundle can be created to validate the Authn Request 
- Add Redis configuration so that it can validate the Message Replay on the Request

### DIFF
--- a/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlApplication.java
+++ b/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlApplication.java
@@ -7,18 +7,34 @@ import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import org.opensaml.core.config.InitializationException;
 import org.opensaml.core.config.InitializationService;
-import uk.gov.ida.dropwizard.logstash.LogstashBundle;
+import se.litsec.opensaml.saml2.common.response.MessageReplayChecker;
 import uk.gov.ida.notification.VerifySamlInitializer;
+import uk.gov.ida.notification.eidassaml.saml.validation.EidasAuthnRequestValidator;
+import uk.gov.ida.notification.eidassaml.saml.validation.components.AssertionConsumerServiceValidator;
+import uk.gov.ida.notification.eidassaml.saml.validation.components.ComparisonValidator;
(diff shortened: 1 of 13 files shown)


JHjava:
@vixus0 We don't. I just removed a whitespace from it.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlApplication.java
```diff
@@ -37,6 +53,33 @@ public void initialize(final Bootstrap<EidasSamlConfiguration> bootstrap) {
...
+        EidasAuthnRequestValidator eidasAuthnRequestValidator = createEidasAuthnRequestValidator(configuration, connectorMetadataResolverBundle);
+        environment.jersey().register(new EidasSamlResource());
+    }
+
+    private EidasAuthnRequestValidator createEidasAuthnRequestValidator(EidasSamlConfiguration configuration, MetadataResolverBundle hubMetadataResolverBundle) throws Exception {
+        MessageReplayChecker replayChecker = configuration.getReplayChecker().createMessageReplayChecker("eidas-saml-parser");
+        DestinationValidator destinationValidator = new DestinationValidator(
+                configuration.getProxyNodeAuthnRequestUrl(), configuration.getProxyNodeAuthnRequestUrl().getPath());
+
+        return new EidasAuthnRequestValidator(
```
idrop: We could consider making all these Validators implement a new EidasSAMLValidator interface with a validate(T) method, and have the resource iterate over a supplied list of `EidasSAMLValidator`s.

vixus0: We're basically reimplementing OpenSAML with our validators. What you've mentioned somewhat exists in OpenSAML as `MessageHandler` and `MessageHandlerChain`.

JHjava: Yep, we can change this later on.

PR #56, opened 2019-02-08
Gateway authn request

idrop:
A new proxy node gateway resource, `EidasAuthnRequestResourceV2` to handle eIDAS SAML Authn requests. This will replace `EidasAuthnRequestResource` in a future PR.

The responsibility of `EidasAuthnRequestResourceV2` is to:
* call `eidas-saml-parser` for validating and parsing eIDAS SAML
* call `verify-proxy-node` to build a new Hub SAML payload
* log data returned from these services
* store data in session to be retrieved during Hub response to gateway
* build a webpage to host the Hub SAML form.


proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2.java
```diff
@@ -0,0 +1,107 @@
...
+    @Path("/POST")
+    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+    public View handlePostBinding(
+            @FormParam(SamlFormMessageType.SAML_REQUEST) String encodedEidasAuthnRequest,
+            @FormParam(RelayState.DEFAULT_ELEMENT_LOCAL_NAME) String eidasRelayState,
+            @Session HttpSession session) {
+        return handleAuthnRequest(encodedEidasAuthnRequest, eidasRelayState, session);
+    }
+
+    private View handleAuthnRequest(String encodedEidasAuthnRequest, String eidasRelayState, HttpSession session) {
```
Wynndow: Do we need this method? Could move it all into `handlePostBinding`.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2.java
```diff
@@ -0,0 +1,107 @@
...
+        this.samlFormViewBuilder = samlFormViewBuilder;
+    }
+
+    @GET
+    @Path("/Redirect")
+    public View handleRedirectBinding(
+            @QueryParam(SamlFormMessageType.SAML_REQUEST) String encodedEidasAuthnRequest,
+            @QueryParam("RelayState") String relayState,
+            @Session HttpSession session) {
+        return handlePostBinding(encodedEidasAuthnRequest, relayState, session);
```
Wynndow: Could this call `handleAuthnRequest` instead of `handlePostBinding?`. Save the extra method call.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2.java
Wynndow: Actually, if we call it from the `/Redirect` route, it makes sense to have it.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2.java
```diff
@@ -0,0 +1,107 @@
...
+        return vspService.generateAuthnRequest();
+    }
+
+    private SamlFormView buildSamlFormView(VSPAuthnRequestResponse vspResponse, String relayState) {
+        URI hubUrl = vspResponse.getSsoLocation();
+        String samlRequest = vspResponse.getSamlRequest();
+        return samlFormViewBuilder.buildRequest(hubUrl.toString(), samlRequest, SUBMIT_BUTTON_TEXT, relayState);
+    }
+
+    private void logAuthnRequestInformation(HttpSession session, EidasSamlParserResponse eidasSamlResponse, VSPAuthnRequestResponse vspResponse) {
```
Wynndow: Might be worth changing `eidasSamlResponse` to `eidasSamlParserResponse` just to disambiguate it from the eidas saml response we'll be sending back to the connector later in the journey.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2Test.java
```diff
@@ -0,0 +1,136 @@
...
+    private Handler logHandler;
+
+    @Captor
+    private ArgumentCaptor<LogRecord> captorLoggingEvent;
+
+    @Captor
+    private ArgumentCaptor<EidasSamlParserRequest> captorEidasSamlParserRequest;
+
+    @Before
+    public void setup() {
```
Wynndow: Do we need the teardown for this to remove that logger?

proxy-node-shared/src/main/java/uk/gov/ida/notification/EidasSamlParserService.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.ida.notification;
```
Wynndow: Should this be in shared? It'll only ever be the gateway that uses it - maybe it should live there?

proxy-node-shared/src/main/java/uk/gov/ida/notification/SamlFormViewBuilder.java
```diff
@@ -13,13 +13,17 @@
     public SamlFormView buildRequest(String url, AuthnRequest authnRequest, String submitText, String relayState) {
         String samlMessage = marshaller.transformToString(authnRequest);
         String encodedSamlMessage = Base64.encodeAsString(samlMessage);
+        return buildRequest(url, encodedSamlMessage, submitText, relayState);
+    }
+
+    public SamlFormView buildRequest(String url, String encodedSamlMessage, String submitText, String relayState) {
```
Wynndow: Might be worth adding a unit test to cover this new signature.

proxy-node-shared/src/main/java/uk/gov/ida/notification/dto/EidasSamlParserRequest.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.ida.notification.dto;
```
Wynndow: Same as the service, could this be moved to the gateway rather than in shared?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2.java
idrop: The idea was to emphasise that these resource were equivalent, but yes think it makes sense to just call handleAuthnRequest directly.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2Test.java
idrop: Will add a teardown to remove mock handler from logger

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2.java
```diff
@@ -0,0 +1,107 @@
...
+
+@Path("/SAML2/SSO")
+public class EidasAuthnRequestResourceV2 {
+
+    private final Logger log = Logger.getLogger(getClass().getName());
+    public static final String SESSION_KEY_EIDAS_REQUEST_ID = "eidas_request_id";
+    public static final String SESSION_KEY_EIDAS_CONNECTOR_PUBLIC_KEY = "eidas_connector_public_key";
+    public static final String SESSION_KEY_EIDAS_DESTINATION = "eidas_destination";
+    public static final String SESSION_KEY_HUB_REQUEST_ID = "hub_request_id";
+    public static final String SUBMIT_BUTTON_TEXT = "Post Verify Authn Request to Hub";
```
CharlesIC: If we need to display a page with a button for the users to click that needs to be designed by the content designers. Better still if we could host that page in the hub frontend instead of the gateway - though all that can be done/decided later when we have the new infrastructure in place.

proxy-node-shared/src/main/java/uk/gov/ida/notification/SamlFormViewBuilder.java
idrop: Yes, added unit tests for testing these public methods directly

PR #58, opened 2019-02-08
Implement client for gateway `eidasSamlParserService`

Wynndow:
### EID-1317 Ensure EidasSamlParserResponse fields are not blank  …
We should always have a value for these fields. Using `NotBlank` will
ensure they're not null and have at least one non whitespace character.

### EID-1323 Implement client for EidasSamlParserService  …
This moves the EidasSamlParserService to the gateway as this is the only
place it will be used.

It implements the client for it to call the eidas saml parser, as well
as the configuration.

A service is being generated in the `GatewayApplication` class, ready to
be passed into the new resource when we register it on the environment.

### EID-1317 Check response for errors and validate DTO …
The `NotBlank` annotations on the DTO are not actually checked until you
use a validator on the object.

This also checks the status code on the response, as well as if the
entity is correct (can be processed).

If any of these things fail they'll raise a custom error message. This
can later be handled with a mapper.

### EID-1323 Use jsonClient from ida-utils  …
It handles errors for us and removes some boilerplate.


proxy-node-gateway/src/main/java/uk/gov/ida/notification/services/EidasSamlParserService.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public EidasSamlParserResponse parse(EidasSamlParserRequest eidasSamlParserRequest) {
+        EidasSamlParserResponse response = eidasSamlParserClient.post(eidasSamlParserRequest, eidasSamlParserURI, EidasSamlParserResponse.class);
+
+        return validateEidasSamlResponseParserDTO(response);
+    }
+
+    private EidasSamlParserResponse validateEidasSamlResponseParserDTO(EidasSamlParserResponse eidasSamlParserResponse) {
+        Set<ConstraintViolation<EidasSamlParserResponse>> violations = validator.validate(eidasSamlParserResponse);
```
idrop: feeling a bit uneasy about having to validate the @NotBlank annotations "by hand". Is it possible to use `javax.validation.constraints.Min(1)` or `javax.validation.constraints.Pattern(regex)` instead

Wynndow: Min is used for numbers by the looks of it, but we could used a regex. I'm not sure what the benefit would be though? All it's currently validating is that the values aren't empty strings or null. How do you mean "by hand"? It'd be the same process - annotating the field then calling `validate` on the object?

idrop: You're right, Min is for numbers 🤭. I had thought using an annotation in package `javax.validation.constraints` meant the annotation and field would be evaluated at marshalling time by jersey, rather than us having to create validation code.

Wynndow: Yeah, me too! It doesn't seem to though. If I change the annotations, the validation call will still need to happen I think.

vixus0: Pretty sure Dropwizard takes care of `@NotNull` and `@NotBlank` validations provided you supply `@Valid` as an annotation to the method parameter.

vixus0: Although maybe that's only for Resources :thinking:

Wynndow: Yeah, Phil had a look at this, it looks like `@Valid` works on the server side (server returns a 500 if the object isn't valid) but it doesn't work for validating client side. As this is from a trusted thing, validating client side is probably unnecessary so I'm just going to remove it.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
```diff
@@ -25,6 +26,7 @@ public EidasSamlResource(EidasAuthnRequestValidator eidasAuthnRequestValidator)
     }
 
     @POST
+    @Valid
```
JHjava: If the @Valid annotation here then don't we need to specify annotations within the ResponseDTO itself so the server knows what to validate?

Wynndow: Yep, that's a good point. I'll remove them.

Wynndow:
The JsonClient is a wrapper around a normal jersey client (an error handling one), and a response processor which handles extracting the response json and generating the object you're after from it.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
JHjava: That's cool, just making sure my understanding was correct.

PR #59, opened 2019-02-08
EID-1304 - Validate the AuthnRequest in the ESP

JHjava:
- Add integration tests for the ESP
- Validate the eIDAS Authn Request in the ESP


JHjava:
@idrop Cheers Phil. Yep, I think there is a separate ticket for that

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
```diff
@@ -18,19 +17,23 @@
 @Produces(MediaType.APPLICATION_JSON)
 public class EidasSamlResource {
 
-    public EidasSamlResource() {
+    private EidasAuthnRequestValidator eidasAuthnRequestValidator;
+    private AuthnRequest authnRequest;
```
idrop: Does `private AuthnRequest authnRequest` need to be a field on `EidasSamlResource`?

JHjava: Nope. It can be moved to the post method. I can add that in next PR?

PR #60, opened 2019-02-08
EID-1254: Metadata generator Java app

vixus0:
Generate and sign proxy service and connector node metadata using PKI on the filesystem or using a PKCS11 module (such as a hardware security module). I've made this a separate project from the rest of the services so that we can build it more easily. I'd like to move it out into its own repository at some point.

Probably easiest to review commit-by-commit. All the good stuff is happening in a couple of files, the rest is just test PKI and Gradle nonsense. Interesting files:
* `MetadataGenerator.java`: Where all the meaty work gets done
* `init_softhsm.sh` and `pre-commit.sh`: Tests the code and also useful for setting up your own HSM

Once this is merged, a bit more work needs to be done to get the local PKI stuff to use it.

## Testing

If you have `softhsm2`, `pkcs11-tool` and `xmlsectool` setup in your dev environment, you can just run `./pre-commit.sh`. Otherwise, use `./docker-test.sh` to build and run tests from a Docker container.


mdgen/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-bin.zip
```
tomrosier: We should upgrade to the latest, `5.2.1`

PR #61, opened 2019-02-12
store RelayState sent by country in order to send back with eiDAS SAML Response

idrop:
A small PR to store the RelayState sent by the country as part of its Authn request.
The RelayState is stored in the http session, and logged.


idrop:
Closing; will merge into integration PR

PR #63, opened 2019-02-12
Eid 1328 - Implement gateway client for calling VSP

Wynndow:
### EID-1328 Add generateAuthnRequest to VSPProxy  …
This adds a method in the VerifyServiceProvider proxy to call the
generateAuthnRequest endpoint of the VSP.

It also adds a method on the VerifyServiceProviderConfiguration to build
the proxy. This is to reduce boilerplate in the apps main Application
class.

It also consolidates all the shared dtos into one package, contracts,
rather than being split over two.

Finally it updates the VSP dtos with the necessary constructors.

### EID-1328 Use vspProxy builder in Translator  …
As the VerifyServiceProviderConfiguraton now has a builder for the
VerifyServiceProviderProxy we can reduce boilerplate in the translators
main Application class.

### EID-1328 Implement VerifyServiceProxy in gateway  …
This pulls in the VSPProxy from shared and replaces the VSPService that
was being used a stub already.

### EID-1328 Rename EidasSamlParserService to EidasSamlParserProxy  …
This is to remain consistent with the VerifyServiceProviderProxy in
shared.


proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceV2.java
```diff
@@ -34,16 +34,16 @@
     public static final String SESSION_KEY_HUB_REQUEST_ID = "hub_request_id";
     public static final String SUBMIT_BUTTON_TEXT = "Post Verify Authn Request to Hub";
 
-    private final EidasSamlParserService eidasSamlParserService;
-    private final VSPService vspService;
+    private final EidasSamlParserProxy eidasSamlParserService;
+    private final VerifyServiceProviderProxy vspProxy;
     private final SamlFormViewBuilder samlFormViewBuilder;
 
     public EidasAuthnRequestResourceV2(
```
dbes-gds: Can we remove the V2 later on when everything is working and delete the old version?

Wynndow: Phil's got a PR which has just been approved which is removing the V2 and just having one resource. We're gonna get this in first then rebase his on top.

PR #68, opened 2019-02-13
EID-1312 - Extract encryption key from connector metadata

JHjava:
- We need to send the encryption key from the connector metadata to the gateway, so extract it. Use the DTO in the VSP for now until the destination and been retrieved from the connector metadata 
- See MetadataTest.java for tests that check the functionally which extracts the public encryption key. I would like to add IntegrationTests for this at some point in the future.

### DIFF
--- a/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/ResponseDto.java
+++ b/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/ResponseDto.java
@@ -10,10 +10,26 @@ public class ResponseDto {
     @JsonProperty
     public String issuer;
 
+    @JsonProperty
+    public String connectorPublicEncryptionKey;
+
     public ResponseDto() {}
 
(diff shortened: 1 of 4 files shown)


eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/ResponseDto.java
```diff
@@ -10,10 +10,26 @@
     @JsonProperty
     public String issuer;
 
+    @JsonProperty
+    public String connectorPublicEncryptionKey;
```
idrop: Could we send back an X509Certificate instead of a String for the cert, as per the shared DTO `EidasSamlParserResponse` i.e.

`@NotNull
    @JsonProperty
    @JsonSerialize(using = X509CertificateSerializer.class)
    @JsonDeserialize(using = X509CertificateDeserializer.class)
    private X509Certificate connectorEncryptionPublicCertificate;`

PR #69, opened 2019-02-13
EID-1321: Pass X509Certificate as a String

CharlesIC:
We can create the cert and credential from the string in the Translator


proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/EidasSamlParserRequest.java
```diff
@@ -7,9 +7,6 @@
     @JsonProperty
     private String authnRequest;
 
-    public EidasSamlParserRequest() {
```
Wynndow: I think the default constructor is needed for Jackson to serialise the DTO.

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/EidasSamlParserResponse.java
```diff
@@ -20,18 +17,13 @@
...
-    @JsonSerialize(using = X509CertificateSerializer.class)
-    @JsonDeserialize(using = X509CertificateDeserializer.class)
-    private X509Certificate connectorEncryptionPublicCertificate;
+    private String connectorEncryptionPublicCertificate;
 
     @JsonProperty
     @NotBlank
     private String destination;
 
-    private EidasSamlParserResponse() {
```
Wynndow: Same as above, I think it's needed for serialisation.

Wynndow:
Why rather a string over the serialized object?

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/EidasSamlParserRequest.java
CharlesIC: I think you're right!

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/EidasSamlParserResponse.java
CharlesIC: Restored 👍

PR #70, opened 2019-02-13
Eid 1297 - Implement hubResponse endpoint in gateway

Wynndow:
### EID-1297 Consolidate URLs to shared  …
The URLs can be used between apps, and only need changing in one place.

### EID-1297 Refactor the Translator proxy  …
This refactors it to use the jsonClient like the other service proxies,
renames it and moves it to the new package.

### EID-1297 Refactor session keys  …
The keys for the gateway session need to be used in the authn request
route as well as the hub response. This defines them in a shared place
so they can be imported as needed.

### EID-1297 Update gateway HubResponseResource  …
The hub response resource now no longer does any validation of the
response, and just forwards it to the translator, before returning the
translated response to the user in a form.

### EID-1297 Delete Apprule tests for HubResponse …
The app rule tests are all out of date and will need to be rewritten.

### EID-1297 Use relayState and destination from session …
The relay state will be stored in the session in the authn request. It's
just a piece of information which the saml standard states should be
returned unaltered, so we just include the one initially sent in the
eidas response.

The destination is confusingly not the destination pulled from the
origin eidas authn request, but the destination the eidas response
should be posted to. The EPS is grabbing this from the connector node
metadata and passing it back to the gateway. We're storing it in the
session in the gateway authn request route.


proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/HubResponseResource.java
```diff
@@ -19,72 +16,60 @@
...
-        }
-    }
+        LOG.info(String.format("[eIDAS Response] received for hub authn request ID '%s', eIDAS authn request ID '%s'", hubRequestId, eidasRequestId));
 
-    private void logHubResponse(HubResponseContainer hubResponseContainer) {
-        LOG.info("[Hub Response] ID: " + hubResponseContainer.getHubResponse().getResponseId());
-        LOG.info("[Hub Response] In response to: " + hubResponseContainer.getHubResponse().getInResponseTo());
-        LOG.info("[Hub Response] Provided level of assurance: " + hubResponseContainer.getAuthnAssertion().getProvidedLoa());
+        return samlFormViewBuilder.buildResponse(
+            connectorNodeUrl,
```
idrop: Could we in future use the AuthnRequest `destination` instead of a static `connectorNodeUrl`?
This will have to be returned from `eidas-saml-parser`.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/HubResponseResource.java
```diff
@@ -19,72 +16,60 @@
...
+            eidasResponse,
+            SUBMIT_TEXT,
+            relayState
```
idrop: NB: In future we should return the `relayState` sent by the country in `AuthnRequest` in preference to the `relayState` send by the Hub. This will mean storing the country's `relayState` in session.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/HubResponseResource.java
Wynndow: I think the destination from the authn request would be the address of the gateway's authn request endpoint. Maybe we could get the ESP to return the issuer - I think that _might_ be the correct url?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/HubResponseResource.java
Wynndow: Yeah that makes senes. I might as well change that as as part of this PR.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/HubResponseResource.java
Wynndow: Actually I've had a look how the issuer is not the same as the endpoint the response should be sent to. The endpoints for current RP's transactions are defined in hub-federation-config, so I think we'll have to do a similar thing here and leave connectorNodeUrl in the config. The URL for the response isn't included in the authn request anywhere.

JHjava: Yep I think that's right. Looking at it this morning, we need the URL which is stored in the connector metadata. We'll extract that in the ESP

Wynndow: James and i just confirmed the connector endpoint that the final response should be posted to can be pulled from the connector metadata. It'll be in the response DTO, currently as `destination`. We should probably change that though as it's not the same `destination` as in the authn request.

PR #73, opened 2019-02-14
Bau - Add Dockerfile for ESP

JHjava:
- Add dockerFile for esp and add to docker-compose
- Change port so it isn't using the same as the gateway

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,12 @@ services:
       context: .
       args:
         component: stub-connector
+  eidas-saml-provider:
+    image: govukverify/eidas-saml-provider:latest
+    build:
+      context: .
+      args:
+        component: eidas-saml-provider
   softhsm:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -18,6 +18,12 @@ services:
       context: .
       args:
         component: stub-connector
+  eidas-saml-provider:
+    image: govukverify/eidas-saml-provider:latest
+    build:
+      context: .
```
blairboy362: At the moment this is not configured to use the Dockerfile that is being added in this PR. It is using the Dockerfile in the root of the tree. We need to either add a `dockerfile` entry (as per line 14) or remove the Dockerfile from this PR and double check that the correct image with the correct binary gets built via the common Dockerfile in the root.

JHjava: Thanks. I'll remove the Dockerfile from the PR so it uses the Dockerfile at root.

PR #74, opened 2019-02-14
Introduce VSP chart

blairboy362:
We're refactoring the behaviour and request journey through our system.
This involves adding two additional apps responsible for validating the
SAML. ESP and VSP will be a crucial part of the system and they deserve
a Chart to define what they look like.

The generation script is somewhat extended to cover some additional
secrets. But it may be broken when it comes to running the setup
locally.


blairboy362:
It's not ready yet.

PR #77, opened 2019-02-14
EID-1322: Return an eIDAS SAML response with correct values

CharlesIC:
Construct a valid SAML response that contains all the required attributes.

  - Add `HubResponseTranslationExceptionMapper`
  - Refactor exception mappers in pn-shared to extend the same base mapper
  - Supply correct values to the hub response translator

### DIFF
--- a/proxy-node-translator/src/dist/config.yml
+++ b/proxy-node-translator/src/dist/config.yml
@@ -16,7 +16,6 @@ proxyNodeResponseUrl: ${PROXY_NODE_RESPONSE_ENDPOINT}
 proxyNodeMetadataForConnectorNodeUrl: ${PROXY_NODE_METADATA_FOR_CONNECTOR_NODE_URL}
 
 hubUrl: ${HUB_URL}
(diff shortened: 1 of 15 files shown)


proxy-node-translator/src/dist/config.yml
```diff
@@ -16,7 +16,6 @@ proxyNodeResponseUrl: ${PROXY_NODE_RESPONSE_ENDPOINT}
 proxyNodeMetadataForConnectorNodeUrl: ${PROXY_NODE_METADATA_FOR_CONNECTOR_NODE_URL}
 
 hubUrl: ${HUB_URL}
```
JHjava: I don't think we need the hubUrl in the translator.

CharlesIC: There's a lot of cleanup to do in the Translator, including the config. We will do the cleanup in a separate story/PR

JHjava: Ok, sounds like a plan.

PR #78, opened 2019-02-14
EID-1333 - Add Connector Metadata healthcheck to ESP

JHjava:
- The ESP will be the only component using the connector metadata, so
it makes sense for it to be running the healthcheck.
- Remove Connector Metadata healtcheck from Gateway
- Add Connector Metadata healthcheck test to ESP

### DIFF
--- a/proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
+++ b/proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
@@ -118,12 +118,6 @@ public class GatewayApplication extends Application<GatewayConfiguration> {
                 environment,
                 "hub-metadata");
 
-        registerMetadataHealthCheck(
-                connectorMetadataResolverBundle.getMetadataResolver(),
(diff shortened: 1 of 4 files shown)


proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
```diff
@@ -118,12 +118,6 @@ public void run(final GatewayConfiguration configuration,
                 environment,
                 "hub-metadata");
 
-        registerMetadataHealthCheck(
-                connectorMetadataResolverBundle.getMetadataResolver(),
```
Wynndow: Does this mean we can remove the connectorMetadataResolverBundle that's instantiated on line 104?

Also, as the gateway isn't using the hub metadata in anyway, can we get rid of the hub metadata health check just above this one?

Wynndow: There's actually quite a lot of stuff in the Application class which can be removed now I think, I'll get a PR up for it.

PR #82, opened 2019-02-15
EID-1332: Adding Helm Charts for ESP.

tomrosier:
This adds the config to get the ESP running in k8's


tomrosier:
Its used as a anchor point for the `./startup.sh` by the looks of it, it uses the names that are listed in that file.

PR #86, opened 2019-02-18
Eid 1313 - Exception handling for gateway authn resource

Wynndow:
The [card for this ticket](https://govukverify.atlassian.net/browse/EID-1313) suggested the gateway should be throwing an `AuthnRequestException` however as the gateway effectively knows nothing about what an authn request is this didn't seem right - that would be the job of the ESP. Instead, I've implemented exception handling for the requests to the two separated micro-services the gateway authn resource relies you.

If the ESP returns a response with a status code in 400/500 range, that status code will be returned to the user. This is because it could have been a mistake on their part and we should let them know. If the VSP returns a non 200 response, we'll always return a 500 as it's totally internal to us with no user input.

These exceptions are mapped to the current error page we have, although the design and content of this will need some work.


proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/mappers/VerifyServiceProviderResponseExceptionMapper.java
```diff
@@ -0,0 +1,20 @@
...
+
+import uk.gov.ida.notification.exceptions.proxy.VerifyServiceProviderResponseException;
+import uk.gov.ida.notification.views.ErrorPageView;
+
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import java.util.logging.Logger;
+
+public class VerifyServiceProviderResponseExceptionMapper implements ExceptionMapper<VerifyServiceProviderResponseException> {
```
CharlesIC: I think we can move this mapper to Gateway since no other component will use it as it returns a PageView. We can have another mapper for VSP errors in the Translator that just returns a 500 with an error message to the Gateway to handle (i.e. show an error page)

Wynndow: I've made the exception and mapper more specific, so only the generate authn endpoint is handled and returned as html. This leaves the translate endpoint free to be mapped in a different way. 
I've had to leave the mapper and exception in shared however. The proxy is part of the shared package which has no access to stuff in the gateway, as it's not a dependency.

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/VerifyServiceProviderProxy.java
```diff
@@ -13,6 +15,7 @@
 
 public class VerifyServiceProviderProxy {
 
+    public static final String LEVEL_OF_ASSURANCE = "LEVEL_2";
```
CharlesIC:
```suggestion
```

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResource.java
```diff
@@ -85,11 +85,11 @@ private void setResponseDataInSession(HttpSession session, EidasSamlParserRespon
...
     }
 
-    private EidasSamlParserResponse parseEidasRequest(String encodedEidasAuthnRequest) {
-        return eidasSamlParserService.parse(new EidasSamlParserRequest(encodedEidasAuthnRequest));
+    private EidasSamlParserResponse parseEidasRequest(String encodedEidasAuthnRequest, String sessionId) {
+        return eidasSamlParserService.parse(new EidasSamlParserRequest(encodedEidasAuthnRequest), sessionId);
     }
 
-    private AuthnRequestResponse generateHubRequestWithVSP() { return vspProxy.generateAuthnRequest(); }
+    private AuthnRequestResponse generateHubRequestWithVSP(String sessionId) { return vspProxy.generateAuthnRequest(sessionId); }
```
CharlesIC: Naming convention note - can we rename this method to `generateHubRequestWithVsp`? We should use consistent camel case naming

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/mappers/VerifyServiceProviderGenerateAuthnRequestResponseExceptionMapper.java
```diff
@@ -0,0 +1,26 @@
...
+
+import uk.gov.ida.notification.exceptions.proxy.VerifyServiceProviderGenerateAuthnRequestResponseException;
+import uk.gov.ida.notification.views.ErrorPageView;
+
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import java.util.logging.Logger;
+
+public class VerifyServiceProviderGenerateAuthnRequestResponseExceptionMapper implements ExceptionMapper<VerifyServiceProviderGenerateAuthnRequestResponseException> {
```
CharlesIC: Maybe rename this to `VspGenerateAuthnRequestResponseExceptionMapper` to make it a manageable length? 😅

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/proxy/VerifyServiceProviderGenerateAuthnRequestResponseException.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.ida.notification.exceptions.proxy;
+
+import javax.ws.rs.WebApplicationException;
+
+public class VerifyServiceProviderGenerateAuthnRequestResponseException extends WebApplicationException {
```
CharlesIC: As above, this could be renamed to `VspGenerateAuthnRequestResponseException`

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/mappers/VerifyServiceProviderResponseExceptionMapper.java
CharlesIC: That's fine 👍

Wynndow:
@CharlesIC All fixed up if you could take another look.

PR #90, opened 2019-02-19
bau - Make esp use shared request class with gateway

JHjava:
- Remove request and response classes from esp which are no longer used
- Change esp resource to use the shared eidas request class as the gateway


eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/base/EidasSamlParserAppRuleTestBase.java
```diff
@@ -65,11 +65,11 @@
             ConfigOverride.config("connectorMetadataConfiguration.trustStore.password", truststore.getPassword())
     );
 
-    protected Response postEidasAuthnRequest(AuthnRequest eidasAuthnRequest) throws MarshallingException {
-        System.out.println(marshaller.transformToString(eidasAuthnRequest));
+    protected Response postEidasAuthnRequest(AuthnRequest authnRequest) throws MarshallingException {
+        System.out.println(marshaller.transformToString(authnRequest));
```
dbes-gds: need this here?

JHjava: Probably not. The tests aren't currently being used so could do with a tidy up when they are.

PR #91, opened 2019-02-20
Eid 1337 - Apprule tests for gateway hub response endpoint

Wynndow:
### EID-1337 Refactor shared URLS  …
Some of the shared URLS were incomplete and hard to work with.

### EID-1337 Add helper for asserting html with XPath  …
When returning html to users we can use xpath to assert we're returning
them the correct thing.

### EID-1337 Refactor gateway authn request app rule tests  …
Previously the tests were dealing with parsing saml, which we don't need
to do - gateway doesn't need to know anything about saml. Instead, we
can test the expected content of the response via xpath.

This also pushes the method for building an authn response up to the
base apprule class as it will be useful for the hub response apprule
tests.

### EID-1337 Refactor test service resources for clientrules  …
The test resources for the VSP and the Translator were both creating
saml objects to be serialized and sent as responses. This is unnecessary
for the gateway as it is unaware of saml. We can instead just provide
strings and make sure they are passed correctly by the gateway.

### EID-1337 Add apprule tests for gateway hub response resource  …
This tests the happy path for the hub response endpoint by firing a
request at it and ensuring the correct values in the html response.

To set this up, we first create a session by hitting the authn endpoint
and capturing the session cookie. This is sent in the request to the hub
response endpoint.

### DIFF
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/Urls.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/Urls.java
@@ -7,8 +7,11 @@ public interface Urls {
     }
 
     interface GatewayUrls {
-        String GATEWAY_EIDAS_AUTHN_REQUEST_PATH = "/SAML2/SSO";
-        String GATEWAY_HUB_RESPONSE_PATH = "/SAML2/SSO/Response";
+        String GATEWAY_ROOT = "/SAML2/SSO";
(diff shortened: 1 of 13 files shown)


proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/Urls.java
```diff
@@ -7,8 +7,11 @@
     }
 
     interface GatewayUrls {
```
tomrosier: We probably can drop `Urls` from this as the parent class is called `Urls`

PR #94, opened 2019-02-20
EID 1320: Translator and Stubbed VSP Tests.

dbes-gds:
Various tests of the Translator /translate-hub-response resource backed-up by a stubbed or mocked VSP Proxy and VerifyServiceProviderProxy /translate-response.

- Testing of the Eidas SAML Response message contents, signing and encryption: HubResponseTranslatorAppRuleTests.java.
- Testing of positive and error cases for VerifyServiceProviderProxy.getTranslatedHubResponse(): VerifyServiceProviderProxyTranslateResponseTest.
- Mock-based testing of /translate-hub-response without an apprule: HubResponseTranslatorTest.  (Basis for further error handling scenarios).
- Some re-factoring of existing VerifyServiceProviderProxy testing to separate /translate-response tests.
- Removal for redundant metadata dependencies from Translator testing.
- Tweaks to DTO to facilitate serialisation.


proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponseTestDataHelper.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.ida.notification.contracts.verifyserviceprovider;
+
+import java.util.Collections;
+
+import org.joda.time.DateTime;
+
+public class TranslatedHubResponseTestDataHelper {
```
JHjava: EidasResponseTestDataGenerator could be a better name? Is eIDASResponse a better name for the TranslatedHubResponse Object too? I guess we only care that we are receiving a eIDAS response back from the VSP.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VerifyServiceProviderProxyGenerateRequestTest.java
```diff
@@ -43,6 +43,7 @@ public AuthnRequestResponse testGenerate(AuthnRequestGenerationBody authnRequest
         }
     }
 
+
```
JHjava: This whitespace could be removed

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VerifyServiceProviderProxyGenerateRequestTest.java
```diff
@@ -125,14 +126,14 @@ public void shouldThrowVerifyServiceProviderResponseExceptionOnClientError() {
...
+                            String.format(
+                                    "Exception of type [CLIENT_ERROR] whilst contacting uri: %s/generate-request",
+                                    testVspClientErrorClientRule.baseUri().toString()
+                            )
+                    );
             assertEquals("session-id", e.getSessionId());
         }
     }
 
-}
```
JHjava: This Whitespace could be removed

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VerifyServiceProviderProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriBuilder;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+import static org.mockito.ArgumentMatchers.eq;
+
+@RunWith(MockitoJUnitRunner.class)
+public class VerifyServiceProviderProxyTranslateResponseTest {
```
JHjava: Long name. Could we shorten it?

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VerifyServiceProviderProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+    public static final DropwizardClientRule testVspClientErrorClientRule = new DropwizardClientRule(new TestVSPClientErrorResource());
+
+    @Spy
+    JsonClient jsonClient = new JsonClient(
+            new ErrorHandlingClient(ClientBuilder.newClient()),
+            new JsonResponseProcessor(new ObjectMapper())
+    );
+
+    @Test
+    public void getTranslatedHubResponseShouldReturnTranslatedHubResponseAuthenticationFailed() {
```
JHjava: Long name again. Could we shorten this?

proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponseTestDataHelper.java
JHjava: I guess the eIDAS response hasn't been generated until it has reached the Translator though.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VerifyServiceProviderProxyTranslateResponseTest.java
CharlesIC: It's a test though - IMO long names in test methods are fine (if ugly!)

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyGenerateRequestTest.java
```diff
@@ -106,12 +106,12 @@ public void shouldThrowVerifyServiceProviderResponseExceptionOnServerError() {
             fail("Expected exception not thrown");
         } catch (VspGenerateAuthnRequestResponseException e) {
             assertThat(e.getCause().getMessage())
-                .startsWith(
-                    String.format(
-                        "Exception of type [REMOTE_SERVER_ERROR] whilst contacting uri: %s/generate-request",
-                        testVspServerErrorClientRule.baseUri().toString()
-                    )
-                );
+                    .startsWith(
```
CharlesIC: When testing for exceptions thrown, AssertJ has a mechanism to capture and check expected properties - these should be used instead of try/catch. Examples:
```
assertThatThrownBy(() -> { throw new Exception("boom!"); }).isInstanceOf(Exception.class)
                                                             .hasMessageContaining("boom");
```
```
assertThatExceptionOfType(IOException.class).isThrownBy(() -> ...
```
```
   // when
   Throwable thrown = catchThrowable(() -> { throw new Exception("boom!"); });

   // then
   assertThat(thrown).isInstanceOf(Exception.class)
```

Guidance here: http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion

Can we please re-write these tests to use the testing framework methods?

CharlesIC: Can disregard the above for this PR as these tests were written previously in a separate PR

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Path("/")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public static class TestVSPResource {
+
+        @POST
+        @Path("/translate-response")
+        public TranslatedHubResponse testGetTranslatedHubResponse(VerifyServiceProviderTranslationRequest request) {
+            if ("SAMLResponse1234".equals(request.getSamlResponse())) {
```
CharlesIC: Can we have two separate resources to make the authn successful vs failed scenarios explicit rather than switching on an arbitrary string value?

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+        TranslatedHubResponse response =
+                vspProxy.getTranslatedHubResponse(
+                        new VerifyServiceProviderTranslationRequest(null, "_1234", "LEVEL_2")
+                );
+
+        assertEquals(VspScenario.AUTHENTICATION_FAILED, response.getScenario());
+        assertEquals("pid1234", response.getPid());
+        assertEquals(VspLevelOfAssurance.LEVEL_2, response.getLevelOfAssurance());
+        assertEquals(null, response.getAttributes());
```
CharlesIC: Can use `assertNull` to have a better error message

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+        assertEquals(VspScenario.IDENTITY_VERIFIED, response.getScenario());
+        assertEquals("123456", response.getPid());
+        assertEquals(VspLevelOfAssurance.LEVEL_2, response.getLevelOfAssurance());
+    }
+
+    @Test(expected = ApplicationException.class)
+    public void shouldThrowApplicationExceptionOnClientError() {
+        VerifyServiceProviderProxy vspProxy = new VerifyServiceProviderProxy(jsonClient, testVspClientErrorClientRule.baseUri());
+
+        TranslatedHubResponse response =
```
CharlesIC: `response` is not used - can remove the variable declaration

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+                        new VerifyServiceProviderTranslationRequest("SAMLResponse1234", "_1234", "LEVEL_2")
+                );
+        fail("Expected exception not thrown");
+    }
+
+    @Test(expected = ApplicationException.class)
+    public void shouldThrowApplicationExceptionOnServerError() {
+        VerifyServiceProviderProxy vspProxy = new VerifyServiceProviderProxy(jsonClient, testVspServerErrorClientRule.baseUri());
+
+        TranslatedHubResponse response =
```
CharlesIC: `response` is not used - can remove the variable declaration

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+    public static final DropwizardClientRule testVspClientErrorClientRule = new DropwizardClientRule(new TestVSPClientErrorResource());
+
+    @Spy
+    JsonClient jsonClient = new JsonClient(
+            new ErrorHandlingClient(ClientBuilder.newClient()),
+            new JsonResponseProcessor(new ObjectMapper())
+    );
+
+    @Test
+    public void shouldReturnTranslatedHubResponseAuthnFailed() {
```
CharlesIC: There are tests to check for the `IDENTITY_VERIFIED` and `AUTHENTICATION_FAILED`
What about tests for `CANCELLATION` and `REQUEST_ERROR`?

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/HubResponseTranslatorTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private final static EidasResponseGenerator eidasResponseGenerator = mock(EidasResponseGenerator.class);
+
+    private final static VerifyServiceProviderProxy verifyServiceProviderProxy = mock(VerifyServiceProviderProxy.class);
+
+    @ClassRule
+    public static final ResourceTestRule resources =
+            ResourceTestRule.builder().addResource(new HubResponseTranslatorResource(eidasResponseGenerator, verifyServiceProviderProxy)).build();
+
+    @Test
+    public void shouldReceive500FromTranslatorWhenEidasResponseNull() {
```
CharlesIC: I'm a bit confused by this test - it doesn't test what it says it does.
The request passed to the resource has a null `samlResponse` but then the VSP call is stubbed to return a valid response - so that cancels out the null SAML response. The test actually fails because the `eidasResponseGenerator.generate` is not stubbed.

In "real life" the VSP would return an error status if passed a null SAML message - perhaps a 200 + `REQUEST_ERROR` - we should test for all possible responses from the VSP rather than guessing, though

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/apprule/HubResponseTranslatorAppRuleTests.java
```diff
@@ -91,78 +97,96 @@ public void setup() throws Throwable {
...
         assertEquals(1, eidasAssertion.getAttributeStatements().size());
         assertEquals(1, eidasAssertion.getAuthnStatements().size());
-
-        assertEquals("Jazzy Harold", attributeStatement.getFirstChild().getTextContent());
     }
 
     @Test
-    @Ignore
-    public void shouldNotAcceptUnsignedHubResponse() throws Exception {
-        javax.ws.rs.core.Response response = postHubResponseToTranslator(buildUnsignedHubResponse());
```
CharlesIC: `buildUnsignedHubResponse` is now unused - can we remove it as well

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
dbes-gds: This keeps the tests for each path on VspProxy separate.  It looked a bit messy when they were both in together.

dbes-gds: Done.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyGenerateRequestTest.java
dbes-gds: Will disregard.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyGenerateRequestTest.java
```diff
@@ -30,7 +30,7 @@
 import static org.mockito.ArgumentMatchers.eq;
 
 @RunWith(MockitoJUnitRunner.class)
-public class VerifyServiceProviderProxyTest {
+public class VSPProxyGenerateRequestTest {
```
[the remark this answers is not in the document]
dbes-gds: VSP changed to Vsp.
There is one test class per path to be tested.  I think this is cleaner than having them all in together.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VerifyServiceProviderProxyTranslateResponseTest.java
dbes-gds: Updated.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
dbes-gds: Done.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
dbes-gds: Removed.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
dbes-gds: Removed.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VSPProxyTranslateResponseTest.java
dbes-gds: Could do, but the behaviour is not actually any different for the other two statuses so the test wouldn't be doing much.

CharlesIC: OK you're right that's not needed for the VSP Proxy test, however, it should be check in a test for either `HubResponseTranslatorResource` or `HubResponseTranslator`

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/HubResponseTranslatorTest.java
dbes-gds: These tests show what happens when there are errors in the Translator resource.  The values passed in aren't really important in this case as that's not what's being tested.  It's a valid scenario for testing though so can add in.

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/HubResponseTranslatorTest.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.UUID;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static uk.gov.ida.saml.core.test.TestCertificateStrings.STUB_COUNTRY_PUBLIC_PRIMARY_CERT;
+
+@RunWith(MockitoJUnitRunner.class)
+public class HubResponseTranslatorTest {
```
CharlesIC: Just spotted this now - can we rename this class to `HubResponseTranslatorResourceTest` - as `HubResponseTranslatorResource` is the class we're testing here.
There is also a class called `HubResponseTranslator` which doesn't have any tests actually - did you think about writing some tests for it as well?

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/apprule/HubResponseTranslatorAppRuleTests.java
dbes-gds: These are deleted lines :)

CharlesIC: Yes, but the deleted code used to call the method `buildUnsignedHubResponse` - after deletion that method is no longer used. Needs to be deleted as well

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/HubResponseTranslatorTest.java
CharlesIC: I was only referring to the `shouldReceive500FromTranslatorWhenEidasResponseNull` test (I had pasted the wrong lines earlier)
It does not test what the test name suggests and in fact it does not test a realistic scenario. At the moment, it's failing because a dependency hasn't been properly mocked - that's not a valid scenario in real usage

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/apprule/HubResponseTranslatorAppRuleTests.java
dbes-gds: OK deleted.

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/HubResponseTranslatorTest.java
CharlesIC: If you want to test unexpected errors in the Translator that's a valid scenario but make it explicit and indicate that in the test name. So, e.g. 
```
 when(eidasTranslator.generate(...)).thenThrow(....)
```

dbes-gds: As the eidasResponseGenerator is not mocked it returns null.  It's not something you would want to happen, but it tests what happens if something unexpected goes wrong.  Maybe it needs a different name?

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/HubResponseTranslatorTest.java
dbes-gds: Ok will do.

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/HubResponseTranslatorTest.java
CharlesIC: Yes, it needs renaming and it would also be better to explicitly simulate a failure/error rather then relying on a side effect of not using a testing framework as it's intended to be used

dbes-gds: Ok I will do something explicit. I did try adding an explicit 'when' to simulate returning a null, however Mockito then gave an error message saying that the mock was unnecessary, because  the default behaviour is actually to return null.

CharlesIC: OK that's fine. We can just explicitly throw an exception - this is what we want to test. And rename the test method to indicate what it's validating (handling unexpected/unhandled/internal errors)

dbes-gds: Removed these tests as Mockito didn't like doing it like this and said the mocks were unnecessary.  Error scenarios are covered by the AppRuleTests which also use the ExceptionMappers.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -87,6 +92,30 @@ public Response translate(HubResponseContainer hubResponseContainer) {
         );
     }
 
+    private void validateHubResponseContainerAttributes(HubResponseContainer hubResponseContainer) {
+        if ( hubResponseContainer == null ) {
+            throw new IllegalArgumentException("HubResponseContainer is null.");
```
JHjava: Would it be better to use a more specific exception? HubResponseException or something like that.

dbes-gds: Updated to use HubResponseTranslationException

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -87,6 +92,30 @@ public Response translate(HubResponseContainer hubResponseContainer) {
         );
     }
 
+    private void validateHubResponseContainerAttributes(HubResponseContainer hubResponseContainer) {
+        if ( hubResponseContainer == null ) {
+            throw new HubResponseTranslationException("HubResponseContainer is null.");
+        }
+        if (hubResponseContainer.getVspScenario() != null && hubResponseContainer.getVspScenario().equals(VspScenario.IDENTITY_VERIFIED)) {
```
idrop: Could we simplify this to:
`VspScenario.IDENTITY_VERIFIED == hubResponseContainer.getVspScenario()`

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -87,6 +92,30 @@ public Response translate(HubResponseContainer hubResponseContainer) {
         );
     }
 
+    private void validateHubResponseContainerAttributes(HubResponseContainer hubResponseContainer) {
```
CharlesIC: https://github.com/alphagov/verify-proxy-node/blob/cef167e53539d5127dcade5e83d0109654de8afe/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java#L95-L117

This method is unnecessary.  `hubResponseContainer` itself will never be null because we create it ourself in the resource and then pass it on to the Translator. The `Attributes` object itself and all its properties have `@NotNull` flags in the DTOs so they will never be null either by definition.
The only check we need to do from this list is whether DoB has a value (it may be not null but still be missing value):
`hubResponseContainer.getAttributes().getDateOfBirth().getValue() == null`

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/apprule/HubResponseTranslatorAppRuleTests.java
```diff
@@ -91,78 +95,66 @@ public void setup() throws Throwable {
...
-    }
 
-    @Test
-    @Ignore
-    public void shouldNotAcceptUnsignedHubResponse() throws Exception {
-        javax.ws.rs.core.Response response = postHubResponseToTranslator(buildUnsignedHubResponse());
-        String message = response.readEntity(String.class);
-        assertEquals(response.getStatus(), HttpStatus.SC_BAD_REQUEST);
-        assertThat(message).contains("Error handling hub response");
+        TranslatedHubResponseTestAssertions.assertAssertionStatements(decryptedEidasResponse);
```
CharlesIC: It's not clear from the method name `assertAssertionStatements ` what it does. I would call it something like `checkAssertionStatementsCorrect` or `checkAssertionStatementsValid`

proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponseBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public static TranslatedHubResponse buildTranslatedHubResponseIdentityVerifiedNoAttributes() {
+        return new TranslatedHubResponse(VspScenario.IDENTITY_VERIFIED, "123456", VspLevelOfAssurance.LEVEL_2, null);
+    }
+
+    public static TranslatedHubResponse buildTranslatedHubResponseRequestError() {
+        return new TranslatedHubResponse(VspScenario.REQUEST_ERROR, "123456", VspLevelOfAssurance.LEVEL_2, null);
+    }
+
+    public static TranslatedHubResponse buildTranslatedHubResponseOneAttributeOnly() {
```
CharlesIC: It would be better to rename this method to something like `buildTranslatedHubResponseIncompleteAttributes` to indicate the intention

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/saml/HubResponseTranslatorTest.java
```diff
@@ -0,0 +1,165 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void translateShouldReturnResponse() {
```
CharlesIC: Test method not very clear. It can have a name that better explains what is being tested - do we want to test that it returns a correct response given correct attributes or any response at all?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
dbes-gds: At the moment we don't know what the vsp response looks like for the other statuses as we've only seen the happy path.  In the stubbed unit test we can return whatever values we like, so I wanted the contract to be clear if anything is missing rather than just have an unexpected null pointer.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VspProxyTranslateResponseTest.java
```diff
@@ -0,0 +1,157 @@
...
+    public void shouldReturnTranslatedHubResponse() {
+
+        VerifyServiceProviderProxy vspProxy = new VerifyServiceProviderProxy(jsonClient, testVspIdentityVerifiedClientRule.baseUri());
+
+        TranslatedHubResponse response =
+                vspProxy.getTranslatedHubResponse(
+                        new VerifyServiceProviderTranslationRequest("SAMLResponse1234", "_1234", "LEVEL_2")
+                );
+
+        assertEquals(VspScenario.IDENTITY_VERIFIED, response.getScenario());
```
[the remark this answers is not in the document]
dbes-gds: I thought we agreed to do assertJ stuff in another PR?  Seen non assertJ asserts many times.

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/apprule/HubResponseTranslatorAppRuleTests.java
dbes-gds: done

proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponseBuilder.java
dbes-gds: done

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/saml/HubResponseTranslatorTest.java
dbes-gds: name updated to translateShouldReturnValidResponseWhenIdentityVerified

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VspProxyTranslateResponseTest.java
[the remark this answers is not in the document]
CharlesIC: Yes, fine to do this in another PR

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
CharlesIC: The DTOs have `@NotNull` flags so we would never get an unexpected null pointer. In fact, we may need to remove some of those flags later we integrate with a real VSP as we discover when those attributes are null depending on the value of `scenario` attribute. Then we can add appropriate checks but at the moment we don't know what they should be and are not necessary IMO

dbes-gds: I built the validation when the need for it became obvious when writing the unit test.  It may be true when the code is integrated that certain scenarios will never happen, which is good.  But when unit testing it's not always obviously how the code will end up being used.  There are valid unit test scenarios where these can be null, so I think it's good to define what happens if they are null.  The validation will help when we do the integration because it will give you explicit information about what's going wrong, so it makes life easier.  I have raised cards to get clarification on what the data looks like for the other scenarios.  Going through the testing process and thinking about what can go wrong helps to clarify requirements.

PR #96, opened 2019-02-20
Add vsp chart

blairboy362:
Add the chart for the Verify Service Provider to the proxy node.

### DIFF
--- a/pki/generate
+++ b/pki/generate
@@ -101,7 +101,7 @@ hub_config_for_proxy_node = {
   'id' => 'VERIFY-HUB',
   'entity_id' => options.idp_entity_id,
   'assertion_consumer_service_uri' => "#{options.proxy_url}/SAML2/SSO/Response/POST",
-  'organization' => { 'name' => 'Hub', 'display_name' => 'Hub', 'url' => 'http://dev-hub.local' },
+  'organization' => { 'name' => 'Hub', 'display_name' => 'Hub', 'url' => 'https://dev-hub.local' },
   'signing_certificates' => [
     { 'name' => 'hub_signing', 'x509' => strip_pem(idp_signing_keypair.cert.to_pem) }
   ],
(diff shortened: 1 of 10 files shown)


pki/generate
```diff
@@ -172,7 +172,7 @@ Dir.chdir(output_dir) do
 
   proxy_node_vars = {
     'HUB_ENTITY_ID' => options.hub_entity_id,
-    'HUB_URL' => "#{options.idp_url}/stub-idp-demo/SAML2/SSO",
+    'HUB_URL' => "#{options.idp_url}/SAML2/SSO",
```
JHjava: We might want to change idp_url to hub_url in options.rb going forward.

vixus0: We probably shouldn't really be using the script at all...

proxy-node-chart/charts/verify-service-provider/templates/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app: {{ template "verify-service-provider.name" . }}
+    chart: {{ template "verify-service-provider.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: {{ template "verify-service-provider.fullname" . }}
+data:
+  verify-service-provider.yaml: |-
+{{ toYaml .Values.configFile | indent 4 }}
```
paroxp: New line, please and thank you \<insert troll face\>

proxy-node-chart/charts/verify-service-provider/values.yaml
```diff
@@ -0,0 +1,58 @@
+# Default values for verify-service-provider.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: govukverify/verify-service-provider
+  tag: "latest"
```
paroxp: Not a requirement, but we probably should be pinning that in some way or another.

PR #98, opened 2019-02-21
Add the metadata controller chart.

blairboy362:
A chart for the metadata controller.

Solo: @blairboy362

### DIFF
--- a/proxy-node-chart/charts/metadata-controller/templates/customer-ca-crt-configmap.yaml
+++ b/proxy-node-chart/charts/metadata-controller/templates/customer-ca-crt-configmap.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "metadata-controller.fullname" . }}-customer-ca-crt
+  labels:
+    app: {{ template "metadata-controller.name" . }}
+    chart: {{ template "metadata-controller.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  customerCA.crt: {{ required "A valid .Values.customerCaCrt entry required!" .Values.customerCaCrt | b64enc | quote }}
--- a/proxy-node-chart/charts/metadata-controller/templates/statefulset.yaml
+++ b/proxy-node-chart/charts/metadata-controller/templates/statefulset.yaml
@@ -1,0 +1,87 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ include "metadata-controller.fullname" . }}
+  labels:
+    app.kubernetes.io/name: {{ include "metadata-controller.name" . }}
+    helm.sh/chart: {{ include "metadata-controller.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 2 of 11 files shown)


proxy-node-chart/charts/metadata-controller/templates/customer-ca-crt-configmap.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ template "metadata-controller.fullname" . }}
+  labels:
+    app: {{ template "metadata-controller.name" . }}
+    chart: {{ template "metadata-controller.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  hsm_signing_cert.pem: {{ required "A valid .Values.customerCaCrt entry required!" .Values.customerCaCrt | b64enc | quote }}
```
vixus0: You probably don't mean to use `customerCaCrt` here, right? The `hsm_signing_cert` is the cert attached to the private key in the HSM while the `customerCACrt` is used by the cloudhsm_client to access the HSM cluster.

proxy-node-chart/charts/metadata-controller/templates/statefulset.yaml
```diff
@@ -0,0 +1,87 @@
...
+      app.kubernetes.io/name: {{ include "metadata-controller.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "metadata-controller.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      containers:
+      - name: {{ .Chart.Name }}-cloudhsm-client
```
vixus0: As discussed the cloudhsm client should be a separate deployment/pod.

proxy-node-chart/charts/metadata-controller/templates/statefulset.yaml
```diff
@@ -0,0 +1,87 @@
...
+                key: HSM_PASSWORD
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+        volumeMounts:
+          name: hsm-signing-cert
+          mountPath: /etc/verify-proxy-node
+      volumes:
+        - name: customer-ca-crt
```
vixus0: As above, only the cloudhsm client deployment/pod needs the customer CA cert, not the metadata controller.

proxy-node-chart/charts/metadata-controller/templates/statefulset.yaml
chrisfarms: we decided to place cloudhsm client in sibling container afterall

proxy-node-chart/charts/metadata-controller/templates/statefulset.yaml
chrisfarms: we decided to place cloudhsm client in sibling container afterall - so this volume is required for the hsm client again

proxy-node-chart/charts/metadata-controller/templates/customer-ca-crt-configmap.yaml
chrisfarms: I ported the change to the incorrect configmap name + key

PR #102, opened 2019-02-25
EID-1293 - Log hashed details for eidas response

JHjava:
- Once the translator has receivied the response from the VSP,
we should log the hashed attributes as well request id and destination.
- Upgrade saml-libs to latest
- Upgrade to dropwizard 1.3.9

### DIFF
--- a/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
+++ b/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
@@ -1,10 +1,13 @@
 package uk.gov.ida.notification.translator.resources;
 
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
 import org.opensaml.core.xml.io.MarshallingException;
 import org.opensaml.security.SecurityException;
 import org.opensaml.xmlsec.signature.support.SignatureException;
 import uk.gov.ida.common.shared.security.X509CertificateFactory;
 import uk.gov.ida.notification.contracts.HubResponseTranslatorRequest;
+import uk.gov.ida.notification.contracts.verifyserviceprovider.Attribute;
 import uk.gov.ida.notification.contracts.verifyserviceprovider.TranslatedHubResponse;
 import uk.gov.ida.notification.contracts.verifyserviceprovider.VerifyServiceProviderTranslationRequest;
 import uk.gov.ida.notification.saml.SamlObjectMarshaller;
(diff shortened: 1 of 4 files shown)


proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -50,14 +57,40 @@ public Response hubResponse(HubResponseTranslatorRequest hubResponseTranslatorRe
         final X509Certificate encryptionCertificate = X_509_CERTIFICATE_FACTORY.createCertificate(hubResponseTranslatorRequest.getConnectorEncryptionCertificate());
 
         final org.opensaml.saml.saml2.core.Response eidasResponse = eidasResponseGenerator.generate(hubResponseContainer, encryptionCertificate);
+
+        String hashedEidasDetails = hashResponseDetails(
```
vixus0: Are we going to move this specific step using `hashResponseDetails` and the `combineAttributeValues` into a separate class in saml-libs so we can make sure that saml-engine will be doing exactly the same thing?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -50,14 +57,40 @@ public Response hubResponse(HubResponseTranslatorRequest hubResponseTranslatorRe
...
+                translatedHubResponse.getPid(),
+                translatedHubResponse.getAttributes().getFirstName().getValue(),
+                combineAttributeValues(translatedHubResponse.getAttributes().getMiddleNames()),
+                combineAttributeValues(translatedHubResponse.getAttributes().getSurnames()),
+                translatedHubResponse.getAttributes().getDateOfBirth().getValue().format(DateTimeFormatter.ofPattern("YYYY-MM-dd"))
```
tomrosier: This could be brittle

vixus0: What would improve it?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
JHjava: Yes, the hashResponseDetails is in saml-libs now. Moving the combineAttributeValues to saml-libs is a good idea. The proxy-node uses its own object type, so I think this a good thing to do once we have started looking at the Hub.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
JHjava: I've tidied this up and moved onto multiple lines. The DateOfBirth has a @NotNull annotation so null pointers shouldn't be a problem here.

PR #103, opened 2019-02-25
EID-1289: Fix translator deployment chart for cloudhsm

vixus0:
- Remove softhsm stuff from the deployment chart
- Add `SIGNER_CONFIG_TYPE` env var and default it to `cloudhsm`
- Add `HSM_SIGNING_KEY_LABEL` env var to specify the key label for the signing
key in the HSM
- Add a test for the `KeyFileSignerConfiguration`. Need to figure out
how to test the `CloudHsmSignerConfiguration`.


proxy-node-translator/src/test/resources/key_file_signer_config.yml
```diff
@@ -0,0 +1,12 @@
+{
```
tomrosier: This is very Jsony looking YAML

vixus0: :hushed:

PR #107, opened 2019-02-27
EID-1318: Moving session to use Redis

tomrosier:
This moves the sessions from being stored within HTTP sessions to putting them into a Redis Cache.

There has been modifications to how the session validation is being done, moving it into the session object rather than having a separate validator.

There is a in memory version of the 'redis' which is just a hashmap that can be used if there is no server to connect to.


proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResource.java
```diff
@@ -63,24 +66,27 @@ public View handlePostBinding(
             @FormParam(SamlFormMessageType.SAML_REQUEST) String encodedEidasAuthnRequest,
             @FormParam(RelayState.DEFAULT_ELEMENT_LOCAL_NAME) String eidasRelayState,
             @Session HttpSession session) {
```
idrop: Can we use the `relayState` supplied by connector as the redis key for our data? This is provided in the AuthnRequest and returned as the `relayState` in the Hub's response to Proxy Node.
In this way we can remove any dependency on `@Session HttpSession session` which will make the gateway stateless, avoiding a set of bugs around session management.

tomrosier: Sounds good 🙂 lets stick this in a new ticket and discuss it with the team. @JHjava what do you think?

proxy-node-gateway/src/dist/config.yml
```diff
@@ -11,6 +11,8 @@ logging:
   appenders:
     - type: logstash-console
 
+redisServerURI: ${REDIS_SERVER_URI:}
```
Wynndow: Is the trailing `:` important?

tomrosier: Nope, good spot 😄

proxy-node-gateway/src/main/java/uk/gov/ida/notification/redis/SessionCacheManager.java
```diff
@@ -0,0 +1,51 @@
...
+import uk.gov.ida.notification.session.GatewaySessionData;
+
+import java.net.URI;
+
+public class SessionCacheManager implements Managed {
+
+    private StorageType storage;
+
+    public SessionCacheManager(URI redisServerURI) {
+        storage = !redisServerURI.toString().isEmpty() ?
```
Wynndow: Could remove the `!` and flip the ternary options to make it slightly easier to mentally parse?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/redis/exceptions/SessionAlreadyExistsException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.ida.notification.redis.exceptions;
+
+public class SessionAlreadyExistsException extends RuntimeException {
```
Wynndow: Does this need a mapper?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/redis/storage/StorageType.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.ida.notification.redis.storage;
+
+import uk.gov.ida.notification.session.GatewaySessionData;
+
+public interface StorageType {
```
andy-paine: It isn't very clear to me what a `StorageType` should be if I just saw that a class implements it. Maybe something like `SessionStore` would be a bit more reflective of what it is actually doing.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/redis/storage/StorageType.java
```diff
@@ -0,0 +1,12 @@
...
+
+    void addSession(String sessionId, GatewaySessionData sessionData);
+    boolean sessionExists(String sessionId);
+    GatewaySessionData getSession(String sessionId);
+    void start();
```
andy-paine: These are just being handed straight to `Managed` - could we just make this interface `extend Managed` to make it clear that anything that implements this _has_ to be `Managed` but the fundamental behaviour of this interface is about sessions, not stopping and starting?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/session/GatewaySessionData.java
```diff
@@ -30,6 +38,26 @@ public GatewaySessionData(
         this.eidasDestination = eidasSamlParserResponse.getDestination();
         this.eidasRelayState = eidasRelayState;
         this.eidasConnectorPublicKey = eidasSamlParserResponse.getConnectorEncryptionPublicCertificate();
+        validate();
```
andy-paine: When do the `@NotEmpty` validations get run? Feels like this stuff would be better lumped in all together so it checks `@NotEmpty` and runs your custom `validate` together. Don't know how easy that is given the framework is doing all the annotation based validation. Don't get me wrong this is functional, but it would be nice to have a common pattern for validating constructed objects.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/base/GatewayAppRuleTestBase.java
```diff
@@ -50,4 +54,19 @@ protected AuthnRequest buildAuthnRequest() throws Exception {
             .withRandomRequestId()
             .build();
     }
+
+    protected String setupTestRedis() {
+        try {
+            server = RedisServer.newRedisServer();
+            server.start();
+            return "redis://" + server.getHost() + ":" + server.getBindPort() + "/";
```
andy-paine: Dynamic binding port? `Jedis` >>> `embedded redis` already

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceTest.java
```diff
@@ -2,11 +2,13 @@
 
 import com.google.common.collect.Lists;
 import org.apache.commons.lang.StringUtils;
+import org.hibernate.Session;
```
andy-paine: Do you need this?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/HubResponseResourceTest.java
```diff
@@ -130,50 +139,16 @@ public void testsHappyPath() {
...
+        HubResponseResource resource = new HubResponseResource(
             new SamlFormViewBuilder(),
-            translatorProxy
+            translatorProxy,
+            sessionCacheManager
         );
 
-        assertThatThrownBy(() -> { resource.hubResponse("hub_saml_response", "relay_state", session); })
-            .isInstanceOf(SessionAttributeException.class)
-            .hasMessage("eidasConnectorPublicKey field may not be empty, eidasDestination field may not be empty");
```
andy-paine: Do we not care about the message anymore? I think you shouldn't because you are typing all your exceptions but it is worth noting that you are not checking messages (presumably for that reason)

proxy-node-gateway/src/test/java/uk/gov/ida/notification/session/TestRedisSessions.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Mock
+    private GatewaySessionData testData;
+
+    private static final String TEST_KEY = "TEST_KEY";
+
+    @Before
+    public void before() throws IOException, URISyntaxException {
+        server = RedisServer.newRedisServer();
+        server.start();
```
andy-paine: Do you trust your library? 😁 If you do, you don't need to test that lettuce interacts with Redis correctly, all you need to do is test that you are using lettuce correctly. [The hub](https://github.com/alphagov/verify-hub/blob/master/hub/policy/src/test/java/uk/gov/ida/hub/policy/session/RedisSessionStoreTest.java) doesn't spin up Redis at all for unit testing the session store, it just mocks redis out and verifies the correct methods were called.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/redis/SessionCacheManager.java
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.ida.notification.session.GatewaySessionData;
+
+import java.net.URI;
+
+public class SessionCacheManager implements Managed {
+
+    private StorageType storage;
+
+    public SessionCacheManager(URI redisServerURI) {
+        storage = redisServerURI.toString().isEmpty() ?
```
idrop: Small one; think its pretty much impossible to construct a URI that is empty, would a null check suffice?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/redis/SessionRedisCodec.java
```diff
@@ -0,0 +1,47 @@
...
+    private ObjectMapper objectMapper = new ObjectMapper();
+
+    @Override
+    public String decodeKey(ByteBuffer bytes) {
+        return new String(bytes.array(), StandardCharsets.UTF_8);
+    }
+
+    @Override
+    public GatewaySessionData decodeValue(ByteBuffer bytes) {
+        try {
```
idrop: Could we use a try-with-resources block here to auto close the InputStream?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceTest.java
```diff
@@ -116,21 +122,14 @@ private void setupHappyPath() throws URISyntaxException {
...
-
-        verify(session).setAttribute(eq(SESSION_KEY_SESSION_DATA), captorGatewaySessionData.capture());
-        verify(session, times(3)).getId();
+        verify(sessionCacheManager, times(1)).addSession(eq("some session id"), any(GatewaySessionData.class));
+        verify(session, times(1)).getId();
         verify(logHandler, times(7)).publish(captorLoggingEvent.capture());
         verify(eidasSamlParserService).parse(captorEidasSamlParserRequest.capture(), any(String.class));
         verify(vspProxy).generateAuthnRequest(any(String.class));
         verify(samlFormViewBuilder).buildRequest("http://hub.bub", "hub blob", SUBMIT_BUTTON_TEXT, "eidas relay state");
         verifyNoMoreInteractions(vspProxy, eidasSamlParserService, logHandler, samlFormViewBuilder, session);
```
idrop: Can we `verifyNoMoreInteractions` with `sessionCacheManager`?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceTest.java
```diff
@@ -116,21 +122,14 @@ private void setupHappyPath() throws URISyntaxException {
...
     private void verifyHappyPath() {
-        GatewaySessionData expectedSessionData = new GatewaySessionData(
-            eidasSamlParserResponse,
-            vspResponse,
-            "eidas relay state"
-        );
-
-        verify(session).setAttribute(eq(SESSION_KEY_SESSION_DATA), captorGatewaySessionData.capture());
-        verify(session, times(3)).getId();
+        verify(sessionCacheManager, times(1)).addSession(eq("some session id"), any(GatewaySessionData.class));
```
idrop: Tiny: shouldn't need the times(1) here, its implied as once, number of times used can then be verified with verifyNoMoreInteractions(..)

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/HubResponseResourceTest.java
```diff
@@ -87,13 +95,14 @@ public void testsHappyPath() {
             "eidas_relay_state_in_session"
         );
 
-        when(session.getAttribute(SESSION_KEY_SESSION_DATA)).thenReturn(sessionData);
+        when(sessionCacheManager.getSession(ArgumentMatchers.anyString())).thenReturn(sessionData);
```
idrop: Could we use String "session-id" instead of `ArgumentMatchers.anyString()` to make the test more meaningful?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResource.java
JHjava: Good point Phil and yes, if we can be sure RelayState will always be populated. Currently the eidasRelayState in GatewaySessionData isn't specified as NotEmpty. If not RelayState then something else instead of sessionId. We should look into this. I agree with Tom, lets stick this in a different ticket.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EidasAuthnRequestResource.java
```diff
@@ -63,24 +66,27 @@ public View handlePostBinding(
...
 
-    private void setResponseDataInSession(HttpSession session, EidasSamlParserResponse eidasSamlParserResponse, AuthnRequestResponse vspResponse, String eidasRelayState) {
-        GatewaySessionData gatewaySessionData = new GatewaySessionData(
-            eidasSamlParserResponse,
-            vspResponse,
-            eidasRelayState
-        );
-        session.setAttribute(SESSION_KEY_SESSION_DATA, gatewaySessionData);
+    private void setResponseDataInSession(String sessionId, EidasSamlParserResponse eidasSamlParserResponse, AuthnRequestResponse vspResponse, String eidasRelayState) {
+        if (!sessionCacheManager.sessionExists(sessionId)) {
```
JHjava: Wouldn't this always be false? And if not, do we want to just carry on with the flow?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
```diff
@@ -81,10 +84,15 @@ private void registerExceptionMappers(Environment environment) {
...
         environment.jersey().register(new TranslatorResponseExceptionMapper());
         environment.jersey().register(new SessionAttributeExceptionMapper());
+        environment.jersey().register(new SessionMissingExceptionMapper());
+        environment.jersey().register(new SessionAlreadyExistsExceptionMapper());
         environment.jersey().register(new GenericExceptionMapper());
     }
 
     private void registerResources(GatewayConfiguration configuration, Environment environment) {
+
+        SessionCacheManager sessionCacheManager = new SessionCacheManager(configuration.getRedisServerURI());
```
dbes-gds: What happens if it can't connect to the redis server?  Or if the redis url is incorrect?  Does the Gateway not start or do you get some kind of exception?

dbes-gds:
General comment: is there a session timeout somewhere?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/session/GatewaySessionData.java
```diff
@@ -30,6 +38,26 @@ public GatewaySessionData(
         this.eidasDestination = eidasSamlParserResponse.getDestination();
         this.eidasRelayState = eidasRelayState;
         this.eidasConnectorPublicKey = eidasSamlParserResponse.getConnectorEncryptionPublicCertificate();
```
dbes-gds: Presume that Redis can persist certs as Strings without mangling them?  Doesn't need to be base64?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
tomrosier: So the service underneath does not try to connect till its been handed to the managed service, which is line 111, which then calls start which will connect to the service. If it fails to connect then it will throw an exception.

I don't know if this is the behaviour we want or we want it to fall over to the in memory redis.

dbes-gds: A couple of suggestions or things to look out for:
- It should be easy to see from the logs that this is what failed on startup.
- K8 should start the Redis service first?  Or it waits or retries for the Redis service to start?

proxy-node-gateway/src/main/java/uk/gov/ida/notification/session/GatewaySessionData.java
Wynndow: AFAIK, the annotation validations only get run when you call `validate()` on the object and nowhere else. I think that dropwizard will use them when serializing an object as the return of a resource, but outside of that you have to validate yourself.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/HubResponseResourceTest.java
tomrosier: The type of exception being thrown is now different and less likely to be different.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/session/GatewaySessionData.java
tomrosier: Me and @dbes-gds have decided to move this into a new ticket.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/redis/SessionRedisCodec.java
```diff
@@ -0,0 +1,47 @@
...
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+public class SessionRedisCodec implements RedisCodec<String, GatewaySessionData> {
+
+    private ObjectMapper objectMapper = new ObjectMapper();
+
+    @Override
+    public String decodeKey(ByteBuffer bytes) {
+        return new String(bytes.array(), StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
tomrosier: It seems to work 🙌

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/HubResponseAppRuleTests.java
```diff
@@ -189,4 +194,9 @@ private NewCookie getSessionCookie(GatewayAppRule appRule) throws Exception {
         return postEidasAuthnRequest(buildAuthnRequest(), appRule).getCookies().get("gateway-session");
     }
 
+    @AfterAll
```
andy-paine: 🎵 you're my Wonderwall 🎵

dbes-gds: LOL

proxy-node-gateway/src/main/java/uk/gov/ida/notification/session/GatewaySessionData.java
JHjava: Has this ticket been created?

PR #110, opened 2019-02-28
EID-1307 Log authn request attributes in ESP

Wynndow:
### EID-1307 Log authn request attributes in ESP
Cyber needs us to log the attributes of the authn request after it's
been validated by the ESP.

They need the attributes to be key/value pairs in a json log event. We're
using the logstash-console appender which is jsonifying the logs for us.

We can add custom key/value pairs to those logs by adding them to
[logbacks MDC](https://logback.qos.ch/manual/mdc.html).

There will be a follow up piece of work to this to log the hub authn request
in the VSP.

### EID-1307 Refactor gateway authn route attribute logging
This refactors the gateway authn logging to log all the attributes in
one log message. Each attribute will appear as a key/value pair in the
json log event.

### DIFF
--- a/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
+++ b/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
@@ -1,0 +1,24 @@
+package uk.gov.ida.notification.eidassaml.logging;
+
+import org.opensaml.saml.saml2.core.AuthnRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+public class EidasAuthnRequestAttributesLogger {
+    private static final Logger log = LoggerFactory.getLogger(EidasAuthnRequestAttributesLogger.class);
+
+    public static void logAuthnRequestAttributes(AuthnRequest authnRequest) {
+        try {
+            MDC.put("eidasRequestId", authnRequest.getID());
+            MDC.put("eidasDestination", authnRequest.getDestination());
+            MDC.put("eidasIssueInstant", authnRequest.getIssueInstant().toString());
+            MDC.put("eidasIssuer", authnRequest.getIssuer().getValue());
+            log.info("Authn request validated by ESP");
+        } catch (Exception e) {
+            MDC.clear();
+            throw e;
+        }
+        MDC.clear();
+    }
+}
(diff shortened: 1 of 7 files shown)


eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
```diff
@@ -0,0 +1,24 @@
...
+        try {
+            MDC.put("eidasRequestId", authnRequest.getID());
+            MDC.put("eidasDestination", authnRequest.getDestination());
+            MDC.put("eidasIssueInstant", authnRequest.getIssueInstant().toString());
+            MDC.put("eidasIssuer", authnRequest.getIssuer().getValue());
+            log.info("Authn request validated by ESP");
+        } catch (Exception e) {
+            MDC.clear();
+            throw e;
+        }
```
tomrosier: you may be able to use a finally cause for the `MDC.clear()` to save it being repeated.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/logging/CombinedAuthnRequestAttributesLogger.java
```diff
@@ -0,0 +1,40 @@
...
+                    eidasSamlParserResponse.getConnectorEncryptionPublicCertificate(),
+                    10
+                )
+            );
+            MDC.put("hubRequestId", vspResponse.getRequestId());
+            MDC.put("hubUrl", vspResponse.getSsoLocation().toString());
+            log.info("Authn requests received from ESP and VSP");
+        } catch (Exception e) {
+            MDC.clear();
+        }
```
tomrosier: Finally might be more appropriate :)

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
Wynndow: Interesting, I was thinking if there was a better way to do this? How would it look? If an exception is thrown, I do actually want it to throw and interrupt the runtime, I just need to clear the MDC first. If have the `MDS.clear()` in a finally block, won't that let the flow of execution continue?

tomrosier: I think with a finally it does still throw, but it throws then tidies up. 

so it be a bit like,

```
        try {
            MDC.put("eidasRequestId", authnRequest.getID());
            MDC.put("eidasDestination", authnRequest.getDestination());
            MDC.put("eidasIssueInstant", authnRequest.getIssueInstant().toString());
            MDC.put("eidasIssuer", authnRequest.getIssuer().getValue());
            log.info("Authn request validated by ESP");
        } catch (Exception e) {
            throw e;
        } finally {
            MDC.clear();
        }

```
https://stackoverflow.com/questions/4264874/does-a-finally-block-run-even-if-you-throw-a-new-exception seems to give more detail.

Wynndow: Nice, I'll update to that.

PR #116, opened 2019-03-05
EID-1353, Switch JUnit to be AssertJ instead.

tomrosier:
Migrated away from Junit and moved all tests to AssertJ

Updated the libraries aswell while i was at it.

### DIFF
--- a/eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/base/EidasSamlParserAppRuleTestBase.java
+++ b/eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/base/EidasSamlParserAppRuleTestBase.java
@@ -8,7 +8,6 @@ import org.junit.Rule;
 import org.opensaml.core.config.InitializationService;
 import org.opensaml.core.xml.io.MarshallingException;
 import org.opensaml.saml.saml2.core.AuthnRequest;
-import se.litsec.opensaml.utils.ObjectUtils;
 import uk.gov.ida.notification.VerifySamlInitializer;
 import uk.gov.ida.notification.apprule.rules.TestMetadataResource;
 import uk.gov.ida.notification.contracts.EidasSamlParserRequest;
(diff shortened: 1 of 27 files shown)


eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/base/EidasSamlParserAppRuleTestBase.java
```diff
@@ -76,7 +75,7 @@ protected Response postEidasAuthnRequest(AuthnRequest authnRequest) throws Marsh
                     .post(Entity.entity(request, MediaType.APPLICATION_JSON_TYPE));
 
         } catch (URISyntaxException e) {
-            fail(e);
+            fail(e.getMessage());
```
CharlesIC: I don't think we need the `fail`here - the method can `throw` (add to the method signature) this exception and if it does the tests that don't expect the exception will fail naturally.
Also the method will never throw a `MarshallingException ` - this can be removed from the signature.

eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/saml/EidasAuthnRequestValidatorTest.java
```diff
@@ -171,15 +170,11 @@ public void shouldThrowIfAuthnRequestHasProtocolBinding() throws XPathExpression
...
+    @Test(expected = InvalidAuthnRequestException.class)
+    public void shouldThrowIfAuthnRequestHasUnsupportedSamlVersion() throws XPathExpressionException, TransformerException {
         final String expectedMessage = "Bad Authn Request from Connector Node: SAML Version should be " + SAMLVersion.VERSION_20.toString();
 
-        assertThrows(InvalidAuthnRequestException.class, () ->
-                eidasAuthnRequestValidator.validate(eidasAuthnRequestBuilder.withSamlVersion(SAMLVersion.VERSION_10).build()), expectedMessage);
-
-        assertThrows(InvalidAuthnRequestException.class, () ->
-                eidasAuthnRequestValidator.validate(eidasAuthnRequestBuilder.withSamlVersion(SAMLVersion.VERSION_11).build()), expectedMessage);
+        eidasAuthnRequestValidator.validate(eidasAuthnRequestBuilder.withSamlVersion(SAMLVersion.VERSION_10).build());
```
CharlesIC: https://github.com/alphagov/verify-proxy-node/blob/38c164edd019d82f250f168b2d5a02235f5807c1/eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/saml/EidasAuthnRequestValidatorTest.java#L175-L177

`expectedMessage` is now not used - it used to validate the exception message. Also two invalid versions of SAML were tested. This can be more elegantly rewritten using the AssertJ exception validation:
` assertThatThrownBy(() -> { eidasAuthnRequestValidator.validate(...VERSION_10...); }).isInstanceOf(InvalidAuthnRequestException.class).hasMessageContaining("expectedMessage");`

And the same for `VERSION_11` - instead of the `expected = InvalidAuthnRequestException.class` flag

PR #122, opened 2019-03-06
Read attribute value when supplying VSP first name for eIDAS response

idrop:
Read the string value of the firstname attribute. Reinstates two `@Ignored` tests.
A future PR is to tidy up the way we combine first and surnames into the eidas response

### DIFF
--- a/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
+++ b/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
@@ -35,7 +35,7 @@ public class HubResponseTranslator {
     }
 
     private String combineFirstAndMiddleNames(List<Attribute<String>> firstNames, List<Attribute<String>> middleNames) {
-        return firstNames.stream().findFirst().get() + " " + combineAttributeValues(middleNames);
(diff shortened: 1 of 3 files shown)


proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -35,7 +35,7 @@ public HubResponseTranslator(String connectorNodeIssuerId, String proxyNodeMetad
     }
 
     private String combineFirstAndMiddleNames(List<Attribute<String>> firstNames, List<Attribute<String>> middleNames) {
-        return firstNames.stream().findFirst().get() + " " + combineAttributeValues(middleNames);
+        return firstNames.stream().findFirst().get().getValue() + " " + combineAttributeValues(middleNames);
```
chrisholmes: You probably should be checking a few things on the attributes here:
- is it verified?
- is it current?

Also, the middleNames attributes are also provided as historical values so you need to find the 'middleName' that is current and verified. `middleNames` was previously confusingly named and refered to the fact that its value can contained more than one word.

JHjava: Yep, we are planning to do that. Realised yesterday our DTO didn't match up, so put this in as a quick fix to unblock the others.

PR #129, opened 2019-03-12
EID-1366: Validate VSP attributes in Translator

CharlesIC:
Added validation of attributes received from the VSP. The following parameters are checked:
  - No null values
  - `verified` flag is `true`
  - Attributes are current - `validFrom` is in the past and `validTo` is `null` or in the future

Also replaced Junit with AssertJ 

At the moment, the validation happens in the `HubResponseTranslator` and it uses some of the new methods on the `Attribute` object. An alternative would be to have a validator object that would be the only point where we do all validation instead. Opinions welcome.


proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/AttributesBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    private List<Attribute<String>> firstNames = new ArrayList<>(singletonList(createAttribute(FIRST_NAME)));
+    private Attribute<String> middleName = createAttribute(MIDDLE_NAME);
+    private Attribute<String> lastName = createAttribute(LAST_NAME);
+    private Attribute<String> gender = createAttribute(GENDER);
+    private Attribute<DateTime> dateOfBirth = createAttribute(DATE_OF_BIRTH);
+
+    public Attributes build() {
+        return new Attributes(
+                firstNames,
+                middleName != null ? singletonList(middleName) : emptyList(),
```
tomrosier: Could this be a function?

proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/AttributesBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+        return this;
+    }
+
+    public AttributesBuilder withoutGender() {
+        this.gender = null;
+        return this;
+    }
+
+    public static DateTime createDateTime(int year, int month, int day, int hour, int minute) {
+        return new DateTime().withYear(year).withMonthOfYear(month).withDayOfMonth(day).withHourOfDay(hour).withMinuteOfHour(minute);
```
tomrosier: It would be neater to drop each method onto a new line.

vixus0: Could also use `DateTime.parse(String str)` instead.

proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/AttributesBuilder.java
CharlesIC: I don't think that's really needed - I'd leave it as it is

proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/AttributesBuilder.java
CharlesIC: I don't think that's really needed - I'd leave it as it is

PR #131, opened 2019-03-12
EID-1381 Integrate with Snyk

Wynndow:
~~This PR is not ready to be merged. There's a technical issue on Snyk's end. Despite this repo being open source they're still imposing a limit of 200 tests per month on us (opensource projects should be unlimited). I have an ongoing support request with them. The approach and the code can still be reviewed however.~~
Snyk have kindly given us unlimited tests until they've resolved the issue, so this looks like it's ready for review.

### EID-1381 Integrate Snyk CLI with Travis  …
Snyk doesn't play very nicely with gradle sub-projects that use
configurations like we do. As such we can't use the Github integration.
We _can_ use the CLI however with Travis to get a similar result.

We can't test individual sub projects, only the gradle configurations.
The configurations to be tested are defined in one place, `snyk/configurations.sh`
so they can be sourced by the test and monitor scripts to ensure consistency.

If `snyk/test.sh` returns a non-zero exit code, the build will be marked as a failure. 
If any of the configurations show a security issue, the script will return non-zero. At
this point you can update the offending dependency and push to rerun the travis build. 
If the issue can't be fixed, it can be ignored via the snyk UI.

The monitor script is run as part of the `after_success` phase of the
Travis build. This will only be run if all tests pass, as well as the
snyk test. If the monitor script fails for some reason the build won't
be marked as failure. Monitoring means that Snyk will notify us if a new vulnerability
becomes apparent with one of our existing dependencies.

### EID-1381 Update dependencies to appease Snyk  …
Snyk found a few vulnerabilities with our dependencies. Some of these
could be fixed by pulling in newer versions of our own libraries, others
needed constraints on transitive dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,19 @@ language: java
 jdk:
   - openjdk11
 
-script: ./gradlew --no-daemon --quiet test
+install:
+  - npm install -g snyk
+
+script:
+  - ./gradlew --no-daemon --quiet test
+  - snyk/test.sh
 
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+
+after_success:
+  - snyk/monitor.sh
 
 cache:
   directories:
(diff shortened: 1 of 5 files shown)


Wynndow:
🤞

PR #132, opened 2019-03-12
EID-1380 - Log when the Translator signs and what with

JHjava:
- Log when each response is signed and specify whether it is being
signed with the CloudHSM or the KeyFile
- Use MDC.remove() rather than MDC.clear() to preserve state of MDC.
- Further work is needed to tidy up our logging helper classes

### DIFF
--- a/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/configuration/CloudHsmSignerConfiguration.java
+++ b/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/configuration/CloudHsmSignerConfiguration.java
@@ -6,6 +6,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.xml.security.algorithms.JCEMapper;
 import org.opensaml.security.x509.BasicX509Credential;
 import org.opensaml.security.x509.X509Support;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/configuration/CloudHsmSignerConfiguration.java
```diff
@@ -16,12 +18,17 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(CloudHsmSignerConfiguration.class);
+
     @JsonCreator
     public CloudHsmSignerConfiguration(
         @JsonProperty("publicKey") DeserializablePublicKeyConfiguration publicKey,
         @JsonProperty("hsmKeyLabel") String hsmKeyLabel
     ) throws SignerConfigurationException {
         try {
+            LOG.info(String.format("Signing eIDAS Response with Cloud HSM using HSM Key Label %s", hsmKeyLabel));
+            LOG.info(String.format("Signing eIDAS Response with Cloud HSM using Public Key %s",publicKey.getCert()));
```
vixus0: Will this just output the subject name of the cert or will it blat the whole thing into the log? Should check the `toString()` of whatever object `getCert()` returns.

JHjava: I was expecting the whole cert. But it will only happen once on startup. Good point though, I'll check.

PR #133, opened 2019-03-13
EID-1363: Log Hash of Eidas Response Attributes in Proxy Node

dbes-gds:
- Create a string for hashing based on the required Response Attributes.
- Log the hash after creation.
- Unit tests to prove that the pre-hash string is correct based on the agreed rules.
- Hash must be identical to the one created in Hub on the same attributes using the same rules.
- Unit tests to identify and alert when the implementation of the rules has changed (in case this no longer matches the hub implementation).
- Removal of redundant references to Hamcrest libraries.
- Tweaks to AttributesBuilder to support many middleNames, lastNames and datesOfBirth to match-up with the payload returned from the VSP.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,11 +18,10 @@ ext {
     opensaml_version = '3.4.0'
     dropwizard_version = '1.3.9'
     utils_version = '2.0.0-350'
-    saml_libs_version = "${opensaml_version}-176"
+    saml_libs_version = "${opensaml_version}-180"
     soft_hsm_version = '1.1.1'
     build_version = "$opensaml_version-${System.env.BUILD_NUMBER ?: 'SNAPSHOT'}"
 }
-
 
 subprojects {
 
@@ -64,13 +63,13 @@ subprojects {
     }
 
     dependencies {
-        common (
+        common(
                 "javax.xml.bind:jaxb-api:2.3.0",
                 "javax.activation:activation:1.1.1",
                 "io.lettuce:lettuce-core:5.1.4.RELEASE",
         )
 
-        dropwizard (
+        dropwizard(
                 "io.dropwizard:dropwizard-core:${dropwizard_version}",
                 "io.dropwizard:dropwizard-client:${dropwizard_version}",
                 "io.dropwizard:dropwizard-views-mustache:${dropwizard_version}",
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -18,12 +18,11 @@ ext {
     opensaml_version = '3.4.0'
     dropwizard_version = '1.3.9'
     utils_version = '2.0.0-350'
-    saml_libs_version = "${opensaml_version}-176"
+    saml_libs_version = "${opensaml_version}-179"
```
idrop: Think the version is actually 180 for saml-libs, a dropwizard dependency change created 179

build.gradle
```diff
@@ -64,63 +63,63 @@ subprojects {
...
         )
 
-        verify_saml_test (
+        verify_saml_test(
                 "uk.gov.ida:common-test-utils:${utils_version}",
                 "uk.gov.ida:saml-metadata-bindings-test:${saml_libs_version}",
         )
 
-        soft_hsm (
+        soft_hsm(
```
tomrosier: We can delete the soft_hsm now.

dbes-gds: yes, got a separate PR for that coming :)

build.gradle
dbes-gds: updated

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/logging/EidasResponseAttributesHashLoggerHelper.java
```diff
@@ -0,0 +1,47 @@
...
+        eidasResponseAttributesHashLogger.setPid(pid);
+
+        if (attributes != null) {
+            if (attributes.getFirstNames() != null) {
+                attributes.getFirstNames().stream()
+                        .filter(Attribute::isVerified)
+                        .findFirst()
+                        .ifPresent(firstName -> eidasResponseAttributesHashLogger.setFirstName(firstName.getValue()));
+            }
+            if (attributes.getMiddleNames() != null) {
```
idrop: `attributes.getMiddleNames()` is never null so no need for null guard. Preferably also guarantee `getFirstNames()` and other list getters are non null, so this reads easier? Not a deal breaker but makes me happy

dbes-gds: Modified Attributes to guarantee non null and have removed the guards.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -48,20 +44,33 @@ public HubResponseTranslatorResource(EidasResponseGenerator eidasResponseGenerat
...
-        final X509Certificate encryptionCertificate = X_509_CERTIFICATE_FACTORY.createCertificate(hubResponseTranslatorRequest.getConnectorEncryptionCertificate());
+        logResponse(eidasResponse);
 
-        final org.opensaml.saml.saml2.core.Response eidasResponse = eidasResponseGenerator.generate(hubResponseContainer, encryptionCertificate);
+        final EidasResponseAttributesHashLoggerHelper eidasResponseAttributesHashLoggerHelper =
+                new EidasResponseAttributesHashLoggerHelper(EidasResponseAttributesHashLogger.instance());
 
-        logResponseDetailsHash(hubResponseTranslatorRequest, translatedHubResponse);
-        logResponse(eidasResponse);
+        eidasResponseAttributesHashLoggerHelper.applyAttributesToHashLogger(
```
JHjava: We should log the hash before eidasResponseGenerator.generate on line 56.

PR #136, opened 2019-03-13
BAU: Refactor SignerConfiguration into CredentialConfiguration

vixus0:
We want to use the CloudHSM with stub connector too, so this changes the
SignerConfiguration code to just return a BasicX509Credential that we
can use for signing or decryption.

In the meantime I've refactored some stuff around signing and decrypting
in SAML and cleaned up some imports.


proxy-node-shared/src/main/java/uk/gov/ida/notification/configuration/CredentialConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+public abstract class CredentialConfiguration {
+    protected BasicX509Credential credential;
+    protected String algorithm;
+
+    public BasicX509Credential getCredential() { return credential; }
+    public String getAlgorithm() { return algorithm; }
+
+    public void setCredential(BasicX509Credential credential) {
+        this.algorithm = SignatureConstants.ALGO_ID_SIGNATURE_RSA_SHA256;
+        if (credential.getPublicKey() instanceof ECPublicKey) {
```
tomrosier: Could this be a turnery

vixus0: Mebbe, but hopefully this makes it explicit that RSA is the default.

PR #141, opened 2019-03-15
EID-1380 - Log when the Translator signs and specify what it is using

JHjava:
- Log when each response is signed and specify whether it is being
signed with the CloudHSM or the KeyFile
- Use MDC.remove() rather than MDC.clear() to preserve state of MDC

### DIFF
--- a/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
+++ b/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
@@ -15,10 +15,11 @@ public class EidasAuthnRequestAttributesLogger {
             MDC.put("eidasIssueInstant", authnRequest.getIssueInstant().toString());
             MDC.put("eidasIssuer", authnRequest.getIssuer().getValue());
             log.info("Authn request validated by ESP");
-        } catch (Exception e) {
-            throw e;
         } finally {
-            MDC.clear();
+            MDC.remove("eidasRequestId");
+            MDC.remove("eidasDestination");
+            MDC.remove("eidasIssueInstant");
+            MDC.remove("eidasIssuer");
         }
     }
 }
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/SamlObjectSigner.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/SamlObjectSigner.java
@@ -9,6 +9,7 @@ import org.opensaml.xmlsec.SignatureSigningParameters;
 import org.opensaml.xmlsec.signature.support.SignatureException;
 import org.opensaml.xmlsec.signature.support.SignatureSupport;
 import org.opensaml.xmlsec.signature.support.SignatureValidator;
+import uk.gov.ida.notification.logging.SamlSigningLoggerHelper;
 
 public class SamlObjectSigner {
     private final SignatureSigningParameters signingParams;
@@ -17,7 +18,8 @@ public class SamlObjectSigner {
         signingParams = SignatureSigningParametersHelper.build(signingCredential, signingAlgorithm);
     }
 
-    public void sign(SignableSAMLObject signableSAMLObject) throws MarshallingException, SecurityException, SignatureException {
+    public void sign(SignableSAMLObject signableSAMLObject, String responseId) throws MarshallingException, SecurityException, SignatureException {
+        SamlSigningLoggerHelper.logSigningRequest(responseId, signingParams.getSigningCredential().getEntityId());
         SignatureSupport.signObject(signableSAMLObject, signingParams);
         SAMLSignatureProfileValidator signatureProfileValidator = new SAMLSignatureProfileValidator();
         signatureProfileValidator.validate(signableSAMLObject.getSignature());
--- a/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
+++ b/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
@@ -44,7 +44,7 @@ public class HubResponseTranslator {
         this.proxyNodeMetadataForConnectorNodeUrl = proxyNodeMetadataForConnectorNodeUrl;
     }
 
-    Response translate(HubResponseContainer hubResponseContainer) {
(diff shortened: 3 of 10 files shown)


eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
```diff
@@ -18,7 +18,10 @@ public static void logAuthnRequestAttributes(AuthnRequest authnRequest) {
         } catch (Exception e) {
```
dbes-gds: Can remove this, it's redundant.

JHjava: Sure. It was already there but I'll take it out.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
```diff
@@ -15,10 +15,11 @@ public static void logAuthnRequestAttributes(AuthnRequest authnRequest) {
             MDC.put("eidasIssueInstant", authnRequest.getIssueInstant().toString());
             MDC.put("eidasIssuer", authnRequest.getIssuer().getValue());
             log.info("Authn request validated by ESP");
-        } catch (Exception e) {
-            throw e;
         } finally {
-            MDC.clear();
```
Wynndow: What was the issue with `clear()`? I've noticed it's been removed in a few places. Was it removing more than it should?

JHjava: So from my understanding is if there is any MDC requests being processed in parallel, then it would be cleared as well. Specifying remove for each MDC attribute, gives you more certainty that only those will be cleared.

Wynndow: [The docs](https://logback.qos.ch/manual/mdc.html#AdvancedUse) seem to suggest that each thread has its own MDC, so shouldn't be affected. I'm fine with the way this is btw, just trying to understand.

proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/SamlObjectSigner.java
```diff
@@ -17,7 +18,8 @@ public SamlObjectSigner(BasicX509Credential signingCredential, String signingAlg
         signingParams = SignatureSigningParametersHelper.build(signingCredential, signingAlgorithm);
     }
 
-    public void sign(SignableSAMLObject signableSAMLObject) throws MarshallingException, SecurityException, SignatureException {
+    public void sign(SignableSAMLObject signableSAMLObject, String responseId) throws MarshallingException, SecurityException, SignatureException {
+        TranslatorSigningLoggerHelper.logSigningRequest(responseId, signingParams.getSigningCredential().getEntityId());
```
CharlesIC: Since the signing code and the signing logger both live in the shared code, they're not Translator specific. To make it clear, I'd rename `TranslatorSigningLoggerHelper ` to simply `SamlSigningLogger`. The `logSigningRequest ` method will be invoked every time - regardless of where it's called from - and it may not always come the Translator as it is now so just to clear that up.

proxy-node-shared/src/main/java/uk/gov/ida/notification/logging/TranslatorSigningLoggerHelper.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.ida.notification.logging;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+public class TranslatorSigningLoggerHelper {
+
+    private static final Logger log = LoggerFactory.getLogger(TranslatorSigningLoggerHelper.class);
```
CharlesIC: By convention `static final` fields have UPPER_CASE_NAMES

proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/SamlObjectSigner.java
dbes-gds: I tried this and Codacy required Helper on the end of all utility classes.

CharlesIC: As discussed, we can ignore this rule for now. I'll silence it since it doesn't make sense.

And my comment was more about the `Translator` part of the name - the signing method lives in a shared library and can be called from anywhere - not just from the Translator (even if at the moment that happens to be the only places it's used)

JHjava: Happy to change this name.

proxy-node-shared/src/main/java/uk/gov/ida/notification/logging/TranslatorSigningLoggerHelper.java
JHjava: Sure but that's the way it has been done in other logger classes by other devs including yourself, so rather keep it consistent for now.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/logging/EidasAuthnRequestAttributesLogger.java
JHjava: Hmm that looks interesting. I'll have a read of that. Cheers Chris.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -44,7 +44,7 @@ public HubResponseTranslator(
         this.proxyNodeMetadataForConnectorNodeUrl = proxyNodeMetadataForConnectorNodeUrl;
     }
 
-    Response translate(HubResponseContainer hubResponseContainer) {
+    public Response translate(HubResponseContainer hubResponseContainer) {
```
CharlesIC: Why make this method public? It's only accessed in its package hence the default access level of package-private (no modifier)

PR #142, opened 2019-03-15
Eid 1357 - Integration Testing of Redis Gateway Sessions

Wynndow:
This PR may not actually be that useful for reasons explained below. If 
people think it's not worthwhile I'm happy to close it.

### EID-1357 Integration test with embedded redis for sessions  …
It was observed at one point that redis could potentially have an issue
with some special characters in certificates. This seems less risky now as
they are all base64 encoded by the ESP before being sent to the gateway
and stored in session.

This test fires up an embedded redis server, rather than using a mock,
and checks that the certificate is retrieved from the session correctly.

### EID-1357 Use version 5.0.3 of redis  …
The standard version used with embedded-reddis is v2.8.19. We're using
the latest version from Dockerhub which at the time of writing is 5.0.2.

Two binaries have been added, one compiled on a mac and the other
compiled on trusty. Travis runs on Trusty hence the need for both.


proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/EidasAuthnRequestAppRuleTests.java
```diff
@@ -10,16 +10,19 @@
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
+import org.mockito.Spy;
```
CharlesIC: You've got some unused imports. IntelliJ is good at highlighting these and removing unnecessary ones (Code -> Optimise Imports or `alt+cmd+I`). Codacy picks that up as well.

Wynndow: Yeah, I'm trying to commit the change but my yubikey is playing up. Nice intellij tip though, cheers!

build.gradle
```diff
@@ -125,6 +125,8 @@ subprojects {
                 "net.sourceforge.htmlunit:htmlunit:2.28",
                 "com.github.stefanbirkner:system-rules:1.19.0",
                 "com.github.fppt:jedis-mock:0.1.9",
+                "it.ozimov:embedded-redis:0.7.2",
```
tomrosier: Do we need this when we already have `"com.github.fppt:jedis-mock:0.1.9",`?

Wynndow: So the original story was about making sure that redis itself could handle certain characters. Jedis, as far as I can tell, is just a mock for redis and uses a Table under the hood for storage. Embedded redis uses an actual redis executable so in theory would properly test if redis can handle said characters. However, as we're storing certs in base64 it might be a moot point.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/RedisTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Paths;
+
+public class RedisTestRule extends ExternalResource {
+    private Redis redis;
+
+    public RedisTestRule(int port) {
+        File redisServerMac;
+        File redisServerTrusty;
```
tomrosier: One Trusty Boy!

PR #143, opened 2019-03-15
EID-1366: Minor cleanup & refactoring

CharlesIC:
- Reduce code duplication & make some method calls more compact
  - Improve formatting to reduce number of short jagged lines
  - Consolidate Translator loggers 
  - Make sure all dates used in testing are always UTC to avoid failures due to different time zones


proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/apprule/HubResponseTranslatorAppRuleTests.java
```diff
@@ -57,19 +59,23 @@
...
 
+@RunWith(MockitoJUnitRunner.class)
 public class HubResponseTranslatorAppRuleTests extends TranslatorAppRuleTestBase {
 
     private static final String PROXY_NODE_ENTITY_ID = "http://proxy-node.uk";
     private static final String EIDAS_TEST_CONNECTOR_DESTINATION = "http://proxy-node/SAML2/SSO/Response";
     private static final SamlObjectMarshaller MARSHALLER = new SamlObjectMarshaller();
     private static final X509CertificateFactory X_509_CERTIFICATE_FACTORY = new X509CertificateFactory();
 
+    private static final String publicBuild = System.getenv("VERIFY_USE_PUBLIC_BINARIES");
```
tomrosier: Where is this used?

CharlesIC: I don't know - I just moved it, not added it. But I believe this is set by gradle/config/env and is false for prod, true for non-prod

PR #146, opened 2019-03-15
EID-1310: Handle internal errors

CharlesIC:
**Summary of the current state of things**
  - We can’t map all exceptions to SAML error responses because some are thrown before we have the minimal information required to construct and send back a SAML response or when something critical fails (like parsing or signing a SAML message for instance). This can be refactored to be more flexible and allow for more cases to be allowed to return a SAML failure message but this may require changes to the session store class or some other mechanism to pass all the necessary information to `ExceptionToSamlErrorResponseMapper`
  - The current implementation is a bit ugly as we need to call the Translator to generate and sign the message. This also means the mappers need a bunch of objects to be initialised (session store, translator proxy, etc). Maybe this can be refactored to be nicer in a future iteration.
  - We will need an error page after all for the cases when the SAML failure message cannot be generated or fails to generate. We can just re-direct to a static page in frontend.

_Steps correspond to commits - recommend reviewing this by commits_
This PR attempts to return a SAML message whenever possible but further improvements can be made as detailed above.

**Step 1**
Make all exception mappers in `proxy-node-shared` consistent
  - Use Dropwizard's `ErrorMessage` object as response body
  - Have one base mapper that all mappers derive from
  - Ensure every exception is logged before returning a response

**Step 2**
Make all exception mappers in proxy-node-gateway consistent. These mappers currently return an error page to the user. In a later iteration, they will return a page with a SAML form with a failure status and a script to auto-POST it back to the requesting connector node.
  - Have one base mapper that all mappers derive from
  - Ensure every exception is logged before returning a response
  - Log more attributes in base exception mappers
  - Tidy up some tests

**Step 3**
Return an error SAML message whenever possible. We need to the Translator to generate and sign the SAML message.
Not all exception mappers can return a SAML message instead of the error page as some of the exceptions are thrown before the session has been properly initialised. We'd need to refactor a number of things to make this work - something to do in a future iteration.

The SAML message exception mappers need a bunch of stuff to be initialised. In a future iteration we should think of a nicer way to generate SAML error messages.

We should also have a user-presentable version of the error page that we can present to users, or to redirect to a static error page hosted in the frontend.


proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/SessionAlreadyExistsExceptionMapper.java
```diff
@@ -1,7 +1,6 @@
 package uk.gov.ida.notification.exceptions.mappers;
 
 import uk.gov.ida.notification.exceptions.SessionAlreadyExistsException;
-import uk.gov.ida.notification.exceptions.SessionMissingException;
 import uk.gov.ida.notification.views.ErrorPageView;
```
idrop: Can this ExceptionMapper extend BaseExceptionMapper?

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/mappers/BaseExceptionMapper.java
```diff
@@ -18,53 +19,54 @@
...
+
+    protected DateTime getIssueInstant(TException exception) {
+        return null;
     }
 
     private void logException(TException exception) {
         String message = exception.getMessage();
         String logId = String.format("%016x", ThreadLocalRandom.current().nextLong());
 
         LOG.warn(format("Error whilst contacting uri [{0}]; logId: {1}; requestId: {2}; issuer: {3}; issueInstant: {4}; cause: {5}",
```
idrop: Could BaseExceptionMapper have a `Level getLogLevel()` method that returns `warn` but subclasses can override? In that way a subclass could indicate that it wants to log at level warn or info etc.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/SessionAlreadyExistsExceptionMapper.java
CharlesIC: The mappers in proxy-node-gateway will extend another base mapper that will render and post a SAML form, since those handle exception at the point in the journey that interacts with the connector node and the users. The mappers in proxy-node-shared will return errors to the Gateway and we only want them to return JSON 500 responses.

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/mappers/BaseExceptionMapper.java
CharlesIC: I think they should all log at the `error` level tbh. Since these are error that will result in a failure response being sent back to the connector node.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToErrorPageMapper.java
```diff
@@ -0,0 +1,73 @@
...
+
+    protected abstract Response.Status getResponseStatus(TException exception);
+
+    protected abstract String getErrorPageMessage(TException exception);
+
+    String getLogId() {
+        return logId;
+    }
+
+    // TODO: Log these three properties via the logger itself as context attributes
```
JHjava: Can we get rid of this?

JHjava:
Looks good Karol. There's some comments in the code that we can remove though.

CharlesIC:
> Not sure about the `ALLCAPS` variables but otherwise good work 😛

@vixus0 The [Java style guide](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html) has a convention whereby `final static` fields have `UPPER_CASE_NAMES`:

> If your variable stores a constant value, such as static final int NUM_GEARS = 6, the convention changes slightly, capitalizing every letter and separating subsequent words with the underscore character. 

We can decide to ditch this convention if there's a good reason for it but at the moment that's the default

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestTranslatorResource.java
```diff
@@ -2,37 +2,47 @@
...
 import java.util.List;
 
+import static uk.gov.ida.notification.apprule.rules.TestTranslatorClientErrorResource.SAML_ERROR_BLOB;
+
 @Path(Urls.TranslatorUrls.TRANSLATOR_ROOT)
 @Produces(MediaType.APPLICATION_JSON)
 public class TestTranslatorResource {
 
-    private List<HubResponseTranslatorRequest> translatorArgList;
+    public static final String SAML_SUCCESS_BLOB = Base64.encodeAsString("encoded-saml-success-message");
```
tomrosier: ❤️

PR #151, opened 2019-03-20
Add Tracibility and Anti-Affinity ADRs

paroxp:
## What

We're adding some extra complexity into the system, in order to mitigate
some of the risks outlined by NCSC. We should be documenting the
decisions and approaches we have taken, in order to reflect on these in
the future and keep the context alive.

## How to review

- Read through, seek sense

### DIFF
--- a/doc/adr/20190320-additional-tracing.md
+++ b/doc/adr/20190320-additional-tracing.md
@@ -1,0 +1,45 @@
+# Add service mesh for traceability
+
+Date: 2019-03-20
+
+## Status
+
+Accepted
+
+## Context
+
+eIDAS project has been challenged with some security concerns provided by
+National Cyber Security Centre (NCSC). One of these concerns was a possibility
+for a connector application or the connection between given connector and the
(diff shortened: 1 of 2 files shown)


CharlesIC:
FYI rebase off master to fix the Travis build

doc/adr/20190320-additional-tracing.md
```diff
@@ -0,0 +1,45 @@
...
+Sadly, it's out of our control. We do however, have an obligation to monitor
+actions that are going through the system, in order to prove our innocence.
+
+We have made a presumption, that Hardware Security Module (HSM) logs would be
+able to tell us when it has been used to sign a thing and attach it to some of
+the logs that we're currently holding.
+
+This isn't however true for the version of HSM that is provided to us by AWS.
+We have managed to raise a feature request, which will not be coming for a
+while.
```
JHjava: Is it worth adding the date which AWS proposed they would begin work on it, so it isn't completely forgotten about?

PR #159, opened 2019-03-27
EID-1412: Handle internal errors - redirect to static error page

CharlesIC:
Redirect to a static error page hosted in verify-frontend when a failure SAML message cannot be generated and returned to the connector node that sent the authn request.

For the time being, we will redirect to the current error page (hosted by the proxy node) if no redirect URL has been configured. This is to enable the app run locally. In the future, once we have a local env, we will want to make the redirect URL mandatory and have the startup script host a static local error page to redirect to.

### DIFF
--- a/proxy-node-gateway/src/dist/config.yml
+++ b/proxy-node-gateway/src/dist/config.yml
@@ -35,3 +35,5 @@ translatorService:
 redisService:
   local: ${USE_REDIS_LOCAL:-false}
   url: ${REDIS_SERVER_URI}
+
+errorPageRedirectUrl: ${PROXY_NODE_ERROR_PAGE_URI:-}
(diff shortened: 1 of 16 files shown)


proxy-node-gateway/src/dist/config.yml
```diff
@@ -35,3 +35,5 @@ translatorService:
 redisService:
   local: ${USE_REDIS_LOCAL:-false}
   url: ${REDIS_SERVER_URI}
+
+errorPageRedirectUrl: ${PROXY_NODE_ERROR_PAGE_URI:-}
```
JHjava: This will need to be set in the deployment chart for Staging. You should talk to Rafal.

CharlesIC: Yep, that's right. For now this will work but it's set up correctly in staging and can test locally we can remove some more code from PN

PR #160, opened 2019-03-27
BAU: Move logging to SamlObjectSigner

vixus0:
Is there a reason we had it in a separate class?


JHjava:
@vixus0 Mainly because the others were, so followed that convention

vixus0:
Ah ok fair enough, probably a good idea to do a cleanup then.

PR #168, opened 2019-04-17
EID-1384 Tweak acceptance tests

Wynndow:
As part of getting the acceptance tests to run in the release pipeline
it'd be useful if the ones we have ran green. These tests need more work
in general but that's out of scope for this ticket.

The error page we serve from the frontend does not include the http
status code. Also selenium does not currently support access to the
status code of a request, so the error page test has been changed to
just check the content.

### DIFF
--- a/proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
+++ b/proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
@@ -17,7 +17,7 @@ Given("the proxy node is sent a LOA {string} request") do |load_type|
 end
 
 And('they progress through verify') do
-  find('label', :text => 'I’ve used Verify before').click
+  find('label', :text => 'I’ve used GOV.UK Verify before').click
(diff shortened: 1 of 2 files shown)


proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
```diff
@@ -53,8 +53,7 @@
   assert_text('You may have selected the wrong company. Check your emails and text messages for confirmation of who verified you.')
 end
 
-
-Then("the user should be presented with an {string} page") do |http_error_code|
+Then("the user should be presented with an error page") do
   assert_text('Sorry, something went wrong')
-  assert_text('HTTP ' + http_error_code)
+  assert_text('This may be because your session timed out or there was a system error.')
```
CharlesIC: We should be checking also for `Sorry, something went wrong` but we have a separate story to overhaul the acceptance tests so this is fine

PR #171, opened 2019-04-17
BAU: Remove old metadata generation app

CharlesIC:


CharlesIC:
@samcrang Safe to delete this?

PR #179, opened 2019-05-01
Remove opensaml dependencies from gateway

idrop:
the public facing gateway service does not do saml processing so it should not require the opensaml libraries.
These are largely set via transitive dependencies within the `proxy-node-shared` module.

You can check gateway dependencies using:
`./gradlew proxy-node-gateway:dependencies --configuration compile`

This PR removes these dependencies, reducing attack surface area available via the gateway classpath.

### DIFF
--- a/proxy-node-gateway/build.gradle
+++ b/proxy-node-gateway/build.gradle
@@ -1,6 +1,5 @@
 dependencies {
     compile configurations.dropwizard,
-            configurations.opensaml,
             configurations.common,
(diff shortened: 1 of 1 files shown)


CharlesIC:
Well spotted 👍

CharlesIC:
Hmmm there's a few more of these transitive dependencies and some of them are only used for tests - but I didn't manage to only exclude them in non-test compilation

PR #183, opened 2019-05-03
EID-1481 user ProxyNodeLogger in ExceptionErrorPageMapper

JohnWattsGDS:
Remove state from exceptions and instead update MDC with state when exception created.

Simplify ExceptionErrorPageMapper by removing unecessary methods used for logged message since state is no longer contained in exceptions.

Remove unecessary methods from exception mappers derived from ExceptionErrorPageMapper.

Add addContext(Exception) method to ProxyNodeLogger so those details are captured consistently.

Co-authored by: Phil Miller <phillip.miller@digital.cabinet-office.gov.uk>


proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToErrorPageMapper.java
```diff
@@ -1,27 +1,23 @@
...
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 import javax.ws.rs.ext.ExceptionMapper;
 import java.net.URI;
-import java.util.Optional;
-import java.util.concurrent.ThreadLocalRandom;
+import java.util.logging.Level;
 
 import static java.text.MessageFormat.format;
```
CharlesIC: This is unused and can be removed

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/metadata/InvalidMetadataException.java
```diff
@@ -3,6 +3,5 @@
 public class InvalidMetadataException extends RuntimeException {
     public InvalidMetadataException(String message, Exception innerException) {
         super(message);
```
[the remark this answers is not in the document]
idrop: We could propogate the root cause of this exception, and have Dropwizard log this innerException at service startup, which is when this exception would be thrown

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/proxy/VerifyServiceProviderRequestException.java
```diff
@@ -1,14 +1,14 @@
...
+
 import javax.ws.rs.WebApplicationException;
 
 public class VerifyServiceProviderRequestException extends WebApplicationException {
-    private final String sessionId;
 
     public VerifyServiceProviderRequestException(Throwable cause, String sessionId) {
         super(cause);
-        this.sessionId = sessionId;
+        MDC.put(ProxyNodeMDCKey.SESSION_ID.name(), sessionId);
```
idrop: We will _try_ to make the `ProxyNodeLogger` static in a separate PR, and at that time we could replace the call to `MDC.put(...)` with a static call to `ProxyNodeLogger.setContext(...)`

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
```diff
@@ -15,6 +16,15 @@ public void addContext(ProxyNodeMDCKey key, String value) {
         MDC.put(key.name(), value);
     }
 
+    public void addContext(Exception exception) {
+        final String cause = Optional.ofNullable(exception.getCause()).map(Throwable::getMessage).orElse(null);
```
idrop: Can we make this easier to read by replacing with:
`Throwable cause = exception.getCause();
        if (cause != null) {
            addContext(ProxyNodeMDCKey.EXCEPTION_CAUSE, cause.getMessage());
        }`

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/proxy/VerifyServiceProviderRequestException.java
JohnWattsGDS: Agreed.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToErrorPageMapper.java
```diff
@@ -1,27 +1,23 @@
...
 
 public abstract class ExceptionToErrorPageMapper<TException extends Exception> implements ExceptionMapper<TException> {
 
-    private static final Logger LOG = LoggerFactory.getLogger(ExceptionToErrorPageMapper.class);
+    private final ProxyNodeLogger proxyNodeLogger = new ProxyNodeLogger();
```
CharlesIC: I'd say we should pass in an instance of `ProxyNodeLogger` in `GatewayApplication` but this is fine if we're going to try to make `ProxyNodeLogger` static in a future PR

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
JohnWattsGDS: Done.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToErrorPageMapper.java
```diff
@@ -40,34 +36,9 @@ public Response toResponse(TException exception) {
 
     protected abstract Response.Status getResponseStatus(TException exception);
```
CharlesIC: https://github.com/alphagov/verify-proxy-node/blob/35f409399a3c4a204e3bce243d549920c215bd50/proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToErrorPageMapper.java#L37

This method and the overrides can be now removed as they're not used anywhere. Which will make the overrides of `ExceptionToErrorPageMapper` a bit useless but we'll deal with that later....

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToErrorPageMapper.java
```diff
@@ -40,34 +36,9 @@ public Response toResponse(TException exception) {
...
-        final String message = exception.getMessage();
-        final String cause = Optional.ofNullable(exception.getCause()).map(Throwable::getMessage).orElse(null);
-
-        LOG.error(format("Error whilst contacting uri [{0}]; logId: {1}; requestId: {2}; sessionId: {3}, issuer: {4}; issueInstant: {5}; message: {6}, cause: {7}",
-                uriInfo.getPath(), logId, getAuthnRequestId(exception), getSessionId(exception), getIssuerId(exception), getIssueInstant(exception), message, cause),
-                exception
-        );
+        proxyNodeLogger.addContext(exception);
+        // This error level will be addressed in the next PR/Story concerning ExceptionToSamlErrorResponseMapper
+        proxyNodeLogger.log(Level.WARNING, format("Error whilst contacting uri [{0}]", this.uriInfo.getPath()));
```
CharlesIC: What's your plan to deal with this?

CharlesIC:
Please see the comments above. Also you'll need to rebase

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToErrorPageMapper.java
JohnWattsGDS: Not exactly sure at this stage.  At the moment ProxyNodeLogger uses the java.util.logging.Logger which doesn't support ERROR as a level.  We can't just swap out to slf4j style because that uses methods to set the level logged at rather than something being passed in to the log method.  The final plan should involve some discussion across the team as there's no obvious nice solution.

PR #191, opened 2019-05-09
EID-1490: Switch to pulling dependencies from Artifactory

CharlesIC:
For all prod builds we need to ensure that the Proxy Node pulls dependencies from Artifactory instead of Bintray. Travis still needs to access Bintray as it cannot access Artifactory.

  - Added `VERIFY_USE_PUBLIC_BINARIES` flag which is set to `true` for Travis builds
  - If the flag is `false` (default), set `gradle` to pull dependencies from Artifactory
  - Exclude unnecessary transitive dependencies
  - Update libraries to latest versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: java
 
+env:
+  global:
+    - VERIFY_USE_PUBLIC_BINARIES=true
+
 jdk:
   - openjdk11
 
--- a/build.gradle
+++ b/build.gradle
@@ -1,13 +1,19 @@
 buildscript {
     repositories {
-        maven { url 'https://dl.bintray.com/alphagov/maven-test' }
-        jcenter()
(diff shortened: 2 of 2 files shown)


CharlesIC:
Travis is happy ✅

idrop:
The VSP is built in Concourse with `VERIFY_USE_PUBLIC_BINARIES` as `true`.
See [the build pipeline here](https://github.com/alphagov/verify-eidas-pipelines/blob/788cbe3046e2c93b64e096e1ea2706ca189846c3/templates/build.yaml#L353). Is this OK?

CharlesIC:
Well spotted @idrop !
I've set it to `false` for the VSP in this PR: alphagov/verify-eidas-pipelines#41
I've checked and the VSP builds fine using Artifactory

PR #193, opened 2019-05-14
EID-1488 Reduce number of exception mappers

JohnWattsGDS:
Delete unused exceptions.

Reduce exception mappers to one that returns a SAML response form and a base class with two sub classes that return a redirect to the error page (one for exceptions we throw and a generic one for when things really did go wrong).

Minor change to build.gradle to import the GrizzlyWebTestContainerFactory so that rules can inject @Context into the mappers as required during testin


proxy-node-gateway/src/main/java/uk/gov/ida/notification/GatewayApplication.java
```diff
@@ -108,15 +103,10 @@ private void registerExceptionMappers(
...
+        environment.jersey().register(new ErrorPageExceptionMapper(errorPageRedirectUrl));
         environment.jersey().register(new GenericExceptionMapper(errorPageRedirectUrl));
-        environment.jersey().register(new SessionMissingExceptionMapper(errorPageRedirectUrl));
-        environment.jersey().register(new EidasSamlParserResponseExceptionMapper(errorPageRedirectUrl));
-        environment.jersey().register(new FailureResponseGenerationExceptionMapper(errorPageRedirectUrl));
-        environment.jersey().register(new VerifyServiceProviderRequestExceptionMapper(errorPageRedirectUrl));
-
-        environment.jersey().register(new SessionAttributeExceptionMapper(samlFormViewBuilder, translatorProxy, sessionStore));
-        environment.jersey().register(new TranslatorResponseExceptionMapper(samlFormViewBuilder, translatorProxy, sessionStore));
-        environment.jersey().register(new SessionAlreadyExistsExceptionMapper(samlFormViewBuilder, translatorProxy, sessionStore));
```
idrop: Yes! 👍

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestErrorPageExceptionMapperRule.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @ClassRule
+    public static final ResourceTestRule resources = ResourceTestRule.builder()
+                                    .addProvider(new ErrorPageExceptionMapper(getUri(EIDAS_ERROR_PAGE_URI)))
+                                    .addProvider(new GenericExceptionMapper(getUri(GENERIC_ERROR_PAGE_URI)))
+                                    .addResource(new TestExceptionMapperResource())
+                                    .build();
+
+    @Test
+    public void shouldMapEidasSamlParserReponseExceptionToErrorPageExceptionMapper() {
```
CharlesIC:
```suggestion
    public void shouldMapEidasSamlParserResponseExceptionToErrorPageExceptionMapper() {
```

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestErrorPageExceptionMapperRule.java
```diff
@@ -0,0 +1,107 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestErrorPageExceptionMapperRule {
+
+    private static final String EIDAS_ERROR_PAGE_URI = "http://eidas_error_page_uri";
+    private static final String GENERIC_ERROR_PAGE_URI = "http://generic_error_page_uri";
+
+    @ClassRule
+    public static final ResourceTestRule resources = ResourceTestRule.builder()
```
CharlesIC: [By convention](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html), in Java, `static final` variables have ALL_CAPS names

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestErrorPageExceptionMapperRule.java
```diff
@@ -0,0 +1,107 @@
...
+                                    .addProvider(new ErrorPageExceptionMapper(getUri(EIDAS_ERROR_PAGE_URI)))
+                                    .addProvider(new GenericExceptionMapper(getUri(GENERIC_ERROR_PAGE_URI)))
+                                    .addResource(new TestExceptionMapperResource())
+                                    .build();
+
+    @Test
+    public void shouldMapEidasSamlParserReponseExceptionToErrorPageExceptionMapper() {
+        Response response;
+
+        response = resources.getJerseyTest()
```
CharlesIC: This is not essential but all these tests can be slightly refactored to be even simpler.

Since the same client with the same set-up is used in every test, this can be refactored. Add a class field `testClient` and then a `setUp()` method with `@Before`  attribute to initialise it. `@Before` will make the method run before each test. Initialise the client with the common bit in each test (`RESOURCES.getJerseyTest()
                            .client().property(ClientProperties.FOLLOW_REDIRECTS, Boolean.FALSE))`
Then the `response` declaration and assignment can be combined to just `Response response = testClient.target(...)...`

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestExceptionMapperResource.java
```diff
@@ -0,0 +1,90 @@
...
+    public Response getEidasSamlParserResponseException() {
+        throw new EidasSamlParserResponseException(new Exception("This caused an EidasSamlParserResponseException"),
+                                                   SESSION_ID);
+    }
+
+    @GET
+    @Path("/FailureResponseGenerationException")
+    public Response getFailureRepsonseGenerationException() {
+        throw new FailureResponseGenerationException(new Exception("This caused a FailureResponseGenerationException"),
+                                                     EIDAS_REQUEST_ID); //TODO - is this actually an EIDAS_REQUEST_ID?
```
CharlesIC: Well spotted - from checking where the exception is thrown, this is indeed `EIDAS_REQUEST_ID`
We should also change the exception constructor to indicate this clearly and remove `REQUEST_ID` from `ProxyNodeMDCKey` and use `EIDAS_REQUEST_ID` instead in the exception constructor

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
```diff
@@ -0,0 +1,108 @@
+package uk.gov.ida.notification.apprule.rules;
+
+import io.dropwizard.testing.junit.DropwizardClientRule;
```
CharlesIC: There's a lot of unused imports here. Would be good to clean them up.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private boolean checkResponseEntityIsSamlFormResponse(String responseEntity) {
+        // The actual application returns a web page with a form but this stripped down version, returns some JSON
+        // As long as it matches, we can be confident that the correct exception mapper has been used
+        Pattern pattern = Pattern.compile("\\{\"postUrl\":\"postUrl\",\"samlMessageType\":\"samlMessageType\",\"encodedSamlMessage\":\"encodedSamlMessage\",\"submitText\":\"submitText\",\"relayState\":.*\"}");
+        Matcher matcher = pattern.matcher(responseEntity);
+        return matcher.matches();
+    }
+
+    private static URI getUri() {
```
CharlesIC: Is this method used anywhere?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
```diff
@@ -0,0 +1,108 @@
...
+public class TestSamlResponseExceptionMapperRule {
+
+    private static final SamlFormViewBuilder samlFormViewBuilder = mock(SamlFormViewBuilder.class);
+    private static final TranslatorProxy translatorProxy = mock(TranslatorProxy.class);
+    private static final SessionStore sessionStore = mock(SessionStore.class);
+
+    @Before
+    public void before() {
+        when(sessionStore.getSession(any(String.class))).thenReturn(new GatewaySessionData("HubRequestId","EidasRequestId","EidasDestination","EidasConnectorPublicKey","EidasRelayState"));
+        String nullString = null;
```
CharlesIC: This is fine but it's okay to just use `null` as an arg in a method call, especially in a test

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
```diff
@@ -0,0 +1,108 @@
...
+    public static final ResourceTestRule resources = ResourceTestRule.builder()
+                                    .setTestContainerFactory(new GrizzlyWebTestContainerFactory())
+                                    .addProvider(new ExceptionToSamlErrorResponseMapper(samlFormViewBuilder, translatorProxy, sessionStore))
+                                    .addResource(new TestExceptionMapperResource())
+                                    .build();
+
+    @Test
+    public void shouldMapSessionAlreadyExistsExceptionToExceptionToSamlErrorResponseMapper() {
+
+        var response = resources.getJerseyTest()
```
CharlesIC: Fancy Java 10 feature! 
I personally like this but we've not used this before so worth bringing this up with the team so we can decide when it's acceptable to use this.

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/EidasErrorPageException.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.ida.notification.exceptions;
+
+import java.util.logging.Level;
+
+public abstract class EidasErrorPageException extends RuntimeException {
```
CharlesIC: Suggest naming this `ErrorPageException` for consistency with `ErrorPageExceptionMapper`

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/TranslatorResponseException.java
```diff
@@ -1,16 +1,27 @@
...
         super(cause);
         MDC.put(ProxyNodeMDCKey.SESSION_ID.name(), sessionId);
         MDC.put(ProxyNodeMDCKey.HUB_REQUEST_ID.name(), hubRequestId);
         MDC.put(ProxyNodeMDCKey.EIDAS_REQUEST_ID.name(), eidasRequestId);
     }
+
+    @Override
+    public Response.Status getResponseStatus() {
+        ApplicationException cause = (ApplicationException) this.getCause();
+        return cause.getExceptionType() == ExceptionType.CLIENT_ERROR ?
```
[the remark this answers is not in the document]
JohnWattsGDS: No, the constructor has to supply a cause and the only place this is instantiated is in a catch block where the caught exception is is passed in as the cause.  I have however amended the constructor so it takes an ApplicationException to ensure there aren't any class-cast exceptions if it's ever used elsewhere in future.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
JohnWattsGDS: No, it's a left over from when both mappers were being tested in the same class.  It has now been deleted.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
JohnWattsGDS: In this case, passing in null is ambiguous as far as the compiler is concerned as the second argument can be either a String or a Response depending on the overload so you have to use a variable of type String set to null to stop the compiler from getting confused.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
JohnWattsGDS: I forgot I used this.  I remembered it was a thing and wondered if it would work.  It did and I promptly forgot about it.  I've changed it back for now.  I'm 50/50 on whether I like it or not anyway!

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/TranslatorResponseException.java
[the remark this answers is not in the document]
CharlesIC: Sounds good, thanks John 👍

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/ProducesSamlResponseException.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.ida.notification.exceptions;
+
+import javax.ws.rs.core.Response;
+
+public abstract class ProducesSamlResponseException extends RuntimeException {
```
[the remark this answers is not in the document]
JohnWattsGDS: Yep.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/RedisSerializationException.java
```diff
@@ -1,5 +1,5 @@
 package uk.gov.ida.notification.exceptions;
-
+// caught by generic exception mapper
```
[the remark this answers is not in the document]
JohnWattsGDS: It now explicitly maps to the error page ...

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
CharlesIC: Oh, right, fair enough then and good job getting around that!

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/EidasErrorPageException.java
JohnWattsGDS: Yep.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestSamlResponseExceptionMapperRule.java
CharlesIC: I think it's ok in tests but probably not in prod code as it makes it more ambiguous. Feel free to bring this up at a dev meeting or on Slack

proxy-node-gateway/src/main/java/uk/gov/ida/notification/exceptions/mappers/ExceptionToSamlErrorResponseMapper.java
```diff
@@ -71,12 +72,11 @@ public Response toResponse(TException exception) {
         return Response.ok().entity(samlFormView).build();
     }
 
-    protected abstract Response.Status getResponseStatus(TException exception);
-
-    private void logException(TException exception) {
-        final String message = exception.getMessage();
-        final String cause = Optional.ofNullable(exception.getCause()).map(Throwable::getMessage).orElse(null);
+    protected Response.Status getResponseStatus(ProducesSamlResponseException exception) {
```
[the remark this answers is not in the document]
JohnWattsGDS: Yep.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestErrorPageExceptionMapperRule.java
JohnWattsGDS: I followed the convention in all the other @ClassRules (EidasAuthnRequestAppRuleTests.java for example).  I feel like it's okay to not do it in this case as it's not really a 'constant' as such but a thing that's required for all the @ClassRule magic etc. and I'm pretty sure that things like the baseUri() for some of those class rules changes (the port being randomly assigned on spin-up) so ALL_CAPS could be misleading as it implies something will always be the same.

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/rules/TestErrorPageExceptionMapperRule.java
JohnWattsGDS: I've done this which is even simpler (and removes the mental gymnastics of @Before):

```java
    private Response getResponse(String uri) {
        return resources.getJerseyTest()
                        .client().property(ClientProperties.FOLLOW_REDIRECTS, Boolean.FALSE)
                        .target(uri).request().get();
    }
```

CharlesIC: Beautiful 👍

PR #195, opened 2019-05-17
EID-1502: Use ProxyNodeLogger everywhere

CharlesIC:
Make logging more consistent by always using the same logger

Summary of changes:
  - Changes to the ProxyNodeLogger to make it easier and more consistent to use
  - Replace all instances of `java.util.logging` and `SLF4J` loggers with the new ProxyNodeLogger
  - Remove `ProxyNodeLogger` from all constructors as it can be easily instantiated within any class
  - Test MDC params by capturing the log event and extracting the MDC param map

### DIFF
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
@@ -16,7 +16,36 @@ public class ProxyNodeLogger {
         MDC.put(key.name(), value);
     }
 
-    public void addContext(Exception exception) {
+    public void info(String message) {
+        log(Level.INFO, message);
+    }
+
+    public void warning(String message) {
+        log(Level.WARNING, message);
+    }
+
+    public void error(String message) {
+        log(Level.SEVERE, message);
+    }
+
+    public void logException(Exception exception, Level level, String message) {
+        addExceptionContext(exception);
+        log(level, message);
+    }
+
+    private void log(Level level, String message) {
+        logWithCallingFrame(level, () -> message, null);
+    }
+
+    private void log(Level level, String message, Exception exception) {
+        logWithCallingFrame(level, () -> message, exception);
+    }
+
+    private void log(Level level, Supplier<String> message) {
+        logWithCallingFrame(level, message, null);
+    }
+
+    private void addExceptionContext(Exception exception) {
         Throwable cause = exception.getCause();
         if (cause != null) {
             addContext(ProxyNodeMDCKey.EXCEPTION_CAUSE, cause.getMessage());
@@ -26,19 +55,15 @@ public class ProxyNodeLogger {
         addContext(ProxyNodeMDCKey.EXCEPTION_STACKTRACE, ExceptionUtils.getStackTrace(exception));
     }
 
-    public void log(Level level, String message) {
-        logWithCallingFrame(level, () -> message);
-    }
-
-    public void log(Level level, Supplier<String> message) {
-        logWithCallingFrame(level, message);
-    }
-
-    private void logWithCallingFrame(Level level, Supplier<String> message) {
+    private void logWithCallingFrame(Level level, Supplier<String> message, Exception exception) {
         Optional<StackWalker.StackFrame> callingFrame = getCallingStackFrame();
-         callingFrame.ifPresent(f -> {
+        callingFrame.ifPresent(f -> {
             addContext(ProxyNodeMDCKey.LOG_LOCATION, String.format("%s.%s", f.getClassName(), f.getMethodName()));
-            LOG.log(level, message);
+            if (exception != null) {
+                LOG.log(level, message.get(), exception);
+            } else {
+                LOG.log(level, message);
+            }
             MDC.remove(ProxyNodeMDCKey.LOG_LOCATION.name());
         });
     }
--- a/proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/ProxyNodeLoggerTest.java
+++ b/proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/ProxyNodeLoggerTest.java
@@ -1,26 +1,103 @@
 package uk.gov.ida.notification.shared;
 
+import ch.qos.logback.classic.Level;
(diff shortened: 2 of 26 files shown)


proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
```diff
@@ -16,7 +16,36 @@ public void addContext(ProxyNodeMDCKey key, String value) {
...
+
+    private void log(Level level, String message, Exception exception) {
+        logWithCallingFrame(level, () -> message, exception);
+    }
+
+    private void log(Level level, Supplier<String> message) {
+        logWithCallingFrame(level, message, null);
+    }
+
+    private void addExceptionContext(Exception exception) {
```
Wynndow: Where does the context added in the method get removed? Shouldn't we be cleaning up everything from the MDC after we log?

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/ProxyNodeLoggerTest.java
```diff
@@ -1,26 +1,43 @@
...
+        Logger logger = (Logger) LoggerFactory.getLogger(ProxyNodeLogger.class);
+        logger.addAppender(appender);
+
+        proxyNodeLogger.info("test");
+
+        verify(appender).doAppend(loggingEventCaptor.capture());
+
+        final ILoggingEvent logEvent = loggingEventCaptor.getValue();
+        assertThat(logEvent.getMessage()).isEqualTo("test");
+        assertThat(logEvent.getMDCPropertyMap().get(ProxyNodeMDCKey.LOG_LOCATION.name())).doesNotContain(ProxyNodeLogger.class.getName());
```
Wynndow: Should we also assert that `LOG_LOCATION.name()` is equal to what we expect? Otherwise the test still passes if this key is never actually set.

It's probably worth testing some of the other functionality in the PrroxyNodeLogger, like adding exception context,

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
CharlesIC: In the `ProxyNodeLogging` filter:
```
for (ProxyNodeMDCKey key : ProxyNodeMDCKey.values()) {
            MDC.remove(key.name());
        }
```

We clean everything up before sending the response back

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -88,11 +86,12 @@ public Response failureResponse(SamlFailureResponseGenerationRequest failureResp
...
 
+    // TODO: Remove once we set all the headers correctly
     private void logSamlResponse(org.opensaml.saml.saml2.core.Response samlResponse) {
-            proxyNodeLogger.addContext(ProxyNodeMDCKey.EIDAS_REQUEST_ID, Objects.requireNonNullElse(samlResponse.getInResponseTo(), ""));
-            proxyNodeLogger.addContext(ProxyNodeMDCKey.EIDAS_DESTINATION, Objects.requireNonNullElse(samlResponse.getDestination(), ""));
-            proxyNodeLogger.addContext(ProxyNodeMDCKey.EIDAS_ISSUER, samlResponse.getIssuer() != null ? samlResponse.getIssuer().getValue() : "");
-            proxyNodeLogger.log(Level.INFO, EIDAS_RESPONSE_LOGGER_MESSAGE);
-
-        }
+        LOG.addContext(ProxyNodeMDCKey.EIDAS_REQUEST_ID, Objects.requireNonNullElse(samlResponse.getInResponseTo(), ""));
```
Wynndow: How do these values in the MDC get cleaned up after logging?

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
Wynndow: Oh nice.

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
```diff
@@ -26,19 +55,15 @@ public void addContext(Exception exception) {
...
         Optional<StackWalker.StackFrame> callingFrame = getCallingStackFrame();
-         callingFrame.ifPresent(f -> {
+        callingFrame.ifPresent(f -> {
             addContext(ProxyNodeMDCKey.LOG_LOCATION, String.format("%s.%s", f.getClassName(), f.getMethodName()));
-            LOG.log(level, message);
+            if (exception != null) {
+                LOG.log(level, message.get(), exception);
+            } else {
+                LOG.log(level, message);
+            }
```
Wynndow: If we're cleaning up the MDC in the filter, can we remove the line below?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
CharlesIC: In the ProxyNodeLogging filter:

for (ProxyNodeMDCKey key : ProxyNodeMDCKey.values()) {
            MDC.remove(key.name());
        }
We clean everything up before sending the response back

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/ProxyNodeLogger.java
CharlesIC: No, because this key has to be specific for each log event. The others keys are specific to the whole request.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/ProxyNodeLoggerTest.java
CharlesIC: Added more tests 👍

PR #198, opened 2019-05-20
EID-1497 ProxyNode JsonClient to insert JourneyId into headers

JohnWattsGDS:
Provide a JsonClient for the ProxyNode which inserts a JourneyId into headers by default.

### DIFF
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/ProxyNodeJsonClient.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/ProxyNodeJsonClient.java
@@ -1,0 +1,37 @@
+package uk.gov.ida.notification.shared.proxy;
+
+import org.slf4j.MDC;
+import uk.gov.ida.jerseyclient.ErrorHandlingClient;
+import uk.gov.ida.jerseyclient.JsonClient;
+import uk.gov.ida.jerseyclient.JsonResponseProcessor;
+import uk.gov.ida.notification.shared.ProxyNodeMDCKey;
+
+import javax.inject.Inject;
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ProxyNodeJsonClient {
+
+    private final JsonResponseProcessor responseProcessor;
+    private final ErrorHandlingClient errorHandlingClient;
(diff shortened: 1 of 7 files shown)


proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/ProxyNodeJsonClient.java
```diff
@@ -0,0 +1,35 @@
...
+    private final ErrorHandlingClient errorHandlingClient;
+
+    @Inject
+    public ProxyNodeJsonClient(ErrorHandlingClient errorHandlingClient, JsonResponseProcessor responseProcessor) {
+        this.errorHandlingClient = errorHandlingClient;
+        this.responseProcessor = responseProcessor;
+    }
+
+    public <T> T post(Object postBody, URI uri, Class<T> clazz) {
+        return responseProcessor.getJsonEntity(uri, null, clazz, errorHandlingClient.post(uri, getJourneyIdHeader(), postBody));
```
CharlesIC: I think it would be better instead to make `ProxyNodeJsonClient` have in instance of `JsonClient` and proxy to its `post` method so we can abstract all of this away. We don't really want to have a custom implementation to make the HTTP call and handle the response

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/ProxyNodeJsonClient.java
```diff
@@ -0,0 +1,33 @@
...
+        this.jsonClient = new JsonClient(errorHandlingClient, responseProcessor);
+    }
+
+    public <T> T post(Object postBody, URI uri, Class<T> clazz) {
+        return jsonClient.post(uri, null, clazz, getJourneyIdHeader());
+    }
+
+    private Map<String, String> getJourneyIdHeader() {
+        Map<String, String> header = new HashMap<>();
+        header.put(ProxyNodeMDCKey.PROXY_NODE_JOURNEY_ID.name(), MDC.get(ProxyNodeMDCKey.PROXY_NODE_JOURNEY_ID.name()));
```
CharlesIC: Maybe we should log here if the journey ID is not present in MDC... We'll see if this works

PR #200, opened 2019-05-22
EID-1513 Add CloudHSM KeyHandle to logging

JohnWattsGDS:
Use the magic of reflection to call the getHandle method on the CaviumKey because AWS only provide the jar buried inside a CloudHSM client that only runs on Linux or Windows and that's probably a better approach than yanking the jar out and making a package for artifactory so the code builds on our machines.

Pass the key through the code so that it can be logged at the right time.

Co-authored by: Karol Gancarz <karol.gancarz@digital.cabinet-office.gov.uk>


proxy-node-shared/src/main/java/uk/gov/ida/notification/configuration/CloudHsmCredentialConfiguration.java
```diff
@@ -37,9 +38,17 @@ public CloudHsmCredentialConfiguration(
...
+            final Key key = cloudHsmStore.getKey(hsmKeyLabel, null);
             BasicX509Credential credential = new BasicX509Credential(
                     certificate,
-                    (PrivateKey) cloudHsmStore.getKey(hsmKeyLabel, null));
+                    (PrivateKey) key);
+
+            this.keyHandle = (Long) ClassLoader.getSystemClassLoader()
+                                                     .loadClass("com.cavium.key.CaviumKey")
+                                                     .getMethod("getHandle")
+                                                     .invoke(key);
```
CharlesIC: Indentation is a bit off here - in the future, can we please use the IntelliJ formatting to make it consistent

PR #214, opened 2019-06-06
BAU: Make Proxy Node metadata entityID match the URL

CharlesIC:
In eIDAS world the entityID has to match the Metadata URL


andy-paine:
Or do I have to be a special GSP approver?

CharlesIC:
Yeah..... 😞

PR #215, opened 2019-06-10
BAU: Test generating dependency files for GitHub security alerts

CharlesIC:
We can take advantage of the GitHub [security alerts feature](https://help.github.com/en/articles/about-security-alerts-for-vulnerable-dependencies) which sends automated alerts for vulnerabilities detected in project dependencies. This can replace Snyk and is directly integrated with GiHub and enabled for projects using the [dependency graph feature](https://help.github.com/en/articles/listing-the-packages-that-a-repository-depends-on). The dependency graph feature is not compatible with Gradle, however, and requires generating POM files for each project. This is easily done with a Maven plugin for Grade and can be done automatically when building or testing the project locally.

To also ensure that the dependency files stay up-to-date, we can make Travis open a pull request if it detects a stale file on the master branch. We already use the GitHub API for Codacy coverage report and we can use the same credentials to open pull requests.

I want to test this for the Proxy Node to see if this would be useful and whether we can extend it to other Verify repos.

Read more on the GitHub blog: [Introducing new ways to keep your code secure](https://github.blog/2019-05-23-introducing-new-ways-to-keep-your-code-secure/#automated-security-fixes-with-dependabot)


Wynndow:
Looks good! Might be worth moving the script and the poms out of the `snyk` namespace though so if all works well we can just drop the snyk dir?

CharlesIC:
I was planning to test it and then move it elsewhere and get rid of Snyk if it does

CharlesIC:
@Wynndow Moved the dependencies out of `snyk`

pre-commit.sh
```diff
@@ -3,6 +3,11 @@ set -e
 
 ./gradlew clean test
 
+if ! git diff -s --exit-code dependencies/*/pom.xml; then
+  echo '⚠️ Remember to commit and push updated project dependencies files (pom.xml) ⚠️'
```
tomrosier: Seems a redundant message if it adds it automatically on the next line

PR #221, opened 2019-06-20
EID-1530: Add more detail to the Stub Connector response page

CharlesIC:
- Display LOA
  - Display attribute names
  - Display decoded SAML message to easily verify against the eIDAS spec

  - Encrypt assertions in Stub Connector tests
  - Slightly refactor `EidasAssertionBuilder` methods to make it easier to understand where LOA is set

Test Connector response page will look like this:
stub-connector-response-page

### DIFF
--- a/eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/EidasSamlResourceTest.java
+++ b/eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/EidasSamlResourceTest.java
@@ -36,10 +36,9 @@ import static uk.gov.ida.saml.core.test.builders.AuthnRequestBuilder.anAuthnRequ
 @RunWith(MockitoJUnitRunner.class)
 public class EidasSamlResourceTest {
 
-    private final static String TEST_CONNECTOR_DESTINATION = "https://stub_country.acme.eu/stub-country-one/destination";
+    private static final String TEST_CONNECTOR_DESTINATION = "https://stub_country.acme.eu/stub-country-one/destination";
 
     private static EidasAuthnRequestValidator eidasAuthnRequestValidator = mock(EidasAuthnRequestValidator.class);
-
     private static SamlRequestSignatureValidator<AuthnRequest> samlRequestSignatureValidator = mock(SamlRequestSignatureValidator.class);
 
     @ClassRule
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/EidasResponseBuilder.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/EidasResponseBuilder.java
@@ -65,14 +65,18 @@ public class EidasResponseBuilder {
         return this;
     }
 
+    public EidasResponseBuilder addAssertionAttributeStatement(List<Attribute> attributes) {
+        assertionBuilder.addAttributeStatement(attributes);
+        return this;
+    }
+
     public EidasResponseBuilder addAssertionAuthnStatement(String authnStatement, DateTime authnIssueInstant) {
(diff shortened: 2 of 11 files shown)


stub-connector/src/main/java/uk/gov/ida/notification/stubconnector/resources/ReceiveResponseResource.java
```diff
@@ -77,21 +82,22 @@ public ResponseView receiveResponse(
 
         // The eIDAS Response should only contain one Assertion with one AttributeStatement which contains
         // the user's requested attributes
-
-        List<String> attributes = new ArrayList<>();
-
         Response decrypted = decrypter.decrypt(response);
         List<Assertion> assertions = decrypted.getAssertions();
 
+        List<Entry<String, String>> attributesByName = new ArrayList<>();
```
idrop: Any reason this isn't a Map?

CharlesIC: Because Map orders the attributes in alphabetical order by name and then it looks weird onn the result page :trollface:

idrop: Ok, could perhaps use LinkedHashMap but it works as is!

proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/EidasResponseBuilder.java
```diff
@@ -65,14 +65,18 @@ public EidasResponseBuilder withAssertionSubject(String subject) {
         return this;
     }
 
+    public EidasResponseBuilder addAssertionAttributeStatement(List<Attribute> attributes) {
```
Wynndow: It looks like the only place this method is used now is in the new `withLoa` method. Can we just rename it this to `withLoa` rather than calling it from somewhere else?

CharlesIC: I'm assuming you mean the `addAssertionAuthnStatement` method. I left it that way because the LOA statement just happens to also be an AuthnStatement in the eIDAS SAML spec so I made the LOA call the generic element method, so I think that makes sense.

Wynndow: Yeah sorry, that it was I meant. I think it's more complex the way it is. If you were looking at the code you might not have any idea about why it was written that way?

CharlesIC: It makes sense if you look at the eIDAS spec, which this class should reflect IMO

Wynndow: OK, feels like an unnecessary abstraction to me but not a deal breaker.

eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/EidasSamlResourceTest.java
```diff
@@ -36,10 +36,9 @@
 @RunWith(MockitoJUnitRunner.class)
 public class EidasSamlResourceTest {
 
-    private final static String TEST_CONNECTOR_DESTINATION = "https://stub_country.acme.eu/stub-country-one/destination";
+    private static final String TEST_CONNECTOR_DESTINATION = "https://stub_country.acme.eu/stub-country-one/destination";
```
tomrosier: 👍

PR #222, opened 2019-06-21
EID-1508 Log requestLogs with logstash-console

Wynndow:
The app logs being sent to Splunk are all in json format which is
processed nicely by Splunk. The access logs however are not.

Here's an example of a request log next to an app log:
Screen Shot 2019-06-21 at 15 52 51


Wynndow:
Snyk is upset - ignoring issues through the GUI doesn't seem to affect the CLI.

PR #224, opened 2019-06-24
EID-1532 implement correct pid translation

Wynndow:
The eIDAS specification states that a PID should be prefixed with
"{Identifying country nationality code}/{destination country nationality
code}".

This allows a nationality code to be provided via the configuration.
This is a short term solution as it only allows for the proxy node to be
connected to one country.

When we restructure to allow multiple countries this will need
revisiting.

### DIFF
--- a/ci/integration/deploy-pipeline.yaml
+++ b/ci/integration/deploy-pipeline.yaml
@@ -73,6 +73,7 @@ spec:
             CLUSTER_NAME: ((cluster.name))
             CLUSTER_DOMAIN: ((cluster.domain))
             RELEASE_NAME: nl-integration
+            CONNECTOR_NODE_NATIONALITY_CODE: NL
             RELEASE_NAMESPACE: ((namespace-deployer.namespace))
             CLUSTER_PRIVATE_KEY: ((cluster.privateKey))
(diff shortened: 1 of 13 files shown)


ci/integration/deploy-pipeline.yaml
```diff
@@ -100,6 +101,7 @@ spec:
                 --set "connector.entityID=${CONNECTOR_ENTITY_ID}" \
                 --set "connector.metadataURL=${CONNECTOR_METADATA_URL}" \
                 --set "connector.metadataSigningTruststoreBase64=${CONNECTOR_METADATA_SIGNING_TRUSTSTORE_BASE64}" \
+                --set "translator.connectorNodeNationalityCode"=${CONNECTOR_CODE_COUNTRY_CODE} \
```
Wynndow: Why pass `CONNECTOR_CODE_COUNTRY_CODE` as a param to this stage of the deployment instead of hard coding here? Also, elsewhere i've used `NATIONALITY_CODE` to be consistent with the eIDAS spec, might be worth changing here.

CharlesIC: Because Phil said so 🤷‍♂ 
You're right - it should be `NATIONALITY_CODE ` - I've corrected that now

PR #227, opened 2019-06-24
EID-1508: Set requestLog layout to access-json

Wynndow:
`io.dropwizard:dropwizard-json-logging` will need to be made
available via artifactory for the build to go green. I think only an admin
can do that.

Setting the requestLog type to 'logstash-console' on its own, done 
in [PR#222](https://github.com/alphagov/verify-proxy-node/pull/222) didn't
work. In fact, it meant that no request logs at all were shipped. I
think this was due to dropwizard not being able to find the correct
type.

Rather than using the `logstash-console` type we can set the layout of the
requestLogs to json, as suggested by the docs: https://www.dropwizard.io/1.3.12/docs/manual/core.html

To do this we need to pull in the `dropwizard-json-logging` module.

This PR also removes the requestLog changes to the VSP. Due to the
new approach requiring a new library, a change would need to be implemented
and released in the VSP. The json logging lib has been added in alphagov/verify-service-provider#203

If this PR doesn't work (I can't test it locally unfortunately) then the
changes made in [PR#222](https://github.com/alphagov/verify-proxy-node/pull/222) will need reverting.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,5 +1,5 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
-version: v1.13.4
+version: v1.13.5
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
 ignore:
   SNYK-JAVA-COMFASTERXMLJACKSONCORE-450207:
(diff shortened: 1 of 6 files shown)


CharlesIC:
I've added `io/dropwizard/dropwizard-json-logging/**` to `whitelisted-repos`

CharlesIC:
From the [Dropwizard docs](https://www.dropwizard.io/1.3.12/docs/manual/core.html):
```
For request logging:

server:
  requestLog:
    appenders:
      - type: console
        layout:
          type: access-json
```
Don't we want `access-json` instead?

PR #230, opened 2019-06-24
Failed package builds should stick to same release #

idrop:
A failing concourse build at package stage currently requests a new [github-release](https://github.com/concourse/github-release-resource).

Failing package builds will now use the same `release` until they go green.


CharlesIC:
Copied to #240

PR #231, opened 2019-06-25
Disable signing of docker images as latest

smford:
There is an issue when signing docker images that results
in any image signed with the tag "latest" invalidating all
previously signed images.  This is an issue as it prevents
rolling back to a previous release.

This commit disables the signing of tag latest, but should
still result in an image being signed.  Deploying "latest"
should still succeed as the image will be signed any way.

### DIFF
--- a/ci/build/release-pipeline.yaml
+++ b/ci/build/release-pipeline.yaml
@@ -155,7 +155,7 @@ spec:
               build_args:
                 component: proxy-node-gateway
               tag_file: src/.git/short_ref
-              tag_as_latest: true
+              tag_as_latest: false
               tag_prefix: v
           - put: translator-image
             params:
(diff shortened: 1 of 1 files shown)


ci/build/release-pipeline.yaml
```diff
@@ -155,7 +155,7 @@ spec:
               build_args:
                 component: proxy-node-gateway
               tag_file: src/.git/short_ref
-              tag_as_latest: true
+              tag_as_latest: false
```
paroxp: That is the default. I think you could safely take the line out.

It's intention is to have both, custom tag but also tag as latest by the way. I find it especially confusing, as the actual Notarising happens on docker push, meaning both tags should have been appropriately signed.

My only guess, is that the tag is part of the metadata, and the signature breaks on the initial tag? 🤔 Confusing.

CharlesIC:
@chrisfarms has confirmed this is no longer needed - closing

> this should no longer be an issue and can be closed as provenance is not enforced via signing of docker images, but instead by trusting that the chart is signed and that an image digest is in use

PR #232, opened 2019-06-26
EID-1487 Send Istio headers with ProxyNodeJsonClient requests

JohnWattsGDS:
Create an IstioHeaderStorage class that uses ThreadLocal to store Istio headers.  Use a request filter to capture the Istio headers on the way in.  Update ProxyNodeJsonClient so that it has access to an instance of IstioHeaderStorage for retrieving the current request’s Istio headers to pass on.  Append Istio headers to response on the way out and clear the headers out of the thread local.

### DIFF
--- a/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlApplication.java
+++ b/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlApplication.java
@@ -5,6 +5,7 @@ import io.dropwizard.configuration.EnvironmentVariableSubstitutor;
 import io.dropwizard.configuration.SubstitutingSourceProvider;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
 import org.opensaml.core.config.InitializationException;
 import org.opensaml.core.config.InitializationService;
 import org.opensaml.saml.metadata.resolver.MetadataResolver;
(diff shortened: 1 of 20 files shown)


eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlApplication.java
```diff
@@ -157,4 +160,13 @@ private String getX509EncryptionCert(EidasSamlParserConfiguration configuration)
 
         return x509EncryptionCert;
     }
+
+    private void registerInjections(Environment environment) {
+        environment.jersey().register(new AbstractBinder() {
```
tomrosier: Does Inteliji let you do this as a lambda?

JohnWattsGDS: No, it displays it as a lambda but if you type in what it displays, it complains it can't find a method.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/IstioHeaderMapperFilterTest.java
```diff
@@ -26,9 +38,12 @@ public void shouldTransmitIstioHeadersInResponseContext() {
         requestHeaders.put(IstioHeaders.X_B3_SPANID, null);
         when(requestContext.getHeaders()).thenReturn(requestHeaders);
         when(responseContext.getHeaders()).thenReturn(responseHeaders);
-        new IstioHeaderMapperFilter().filter(requestContext, responseContext);
+        doReturn("this isn't being tested").when(istioHeaderStorage).toString();
```
instantiator: I'm not sure I understand why this is here.

JohnWattsGDS: It's actually redundant now.  At some point in a previous iteration, there was a null pointer exception that, that stopped.  It's not the case any more so I've taken it out.  Good spot.

PR #233, opened 2019-06-26
EID-1531: Validate Proxy Node params

instantiator:
- Adds new custom validators
- Updates HubResponseTranslatorRequest for validation
- Updates tests that failed due to new validation requirements
- Updates version of saml-libs to 203


proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/ValidSamlId.java
```diff
@@ -0,0 +1,25 @@
...
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Documented
+@Constraint(validatedBy = { SamlIdValidator.class })
+@Size(min = SamlIdValidator.MinLength, max = SamlIdValidator.MaxLength)
+@Target({ FIELD, PARAMETER })
+@Retention(RUNTIME)
+public @interface ValidSamlId {
```
JohnWattsGDS: I like that you've created annotations for the validation a lot.  Is it possible to be more specific with the message where appropriate so we can see instantly exactly what caused the validation to fail?

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public static final int MinLength = 64;
+    public static final int MaxLength = 64000;
+
+    @Override
+    public void initialize(ValidBase64Saml constraint) { /* intentionally blank */ }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (value == null || StringUtils.isBlank(value)) { return true; } // @NotNull should detect nulls
```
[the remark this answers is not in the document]
instantiator: Will fix 👍

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriStringValidator.java
```diff
@@ -0,0 +1,28 @@
...
+import java.net.URI;
+
+public class DestinationUriStringValidator implements ConstraintValidator<ValidDestinationUriString, String> {
+
+    @Override
+    public void initialize(ValidDestinationUriString constraint) { /* intentionally blank */ }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (StringUtils.isBlank(value)) { return true; } // @NotNull should detect nulls
```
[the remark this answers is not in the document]
instantiator: I've double checked, and isBlank returns true for whitespace, empty string, and null - so it's the one we want here.

proxy-node-translator/src/test/java/uk/gov/ida/notification/translator/validations/HubResponseTranslatorRequestValidationTests.java
```diff
@@ -0,0 +1,77 @@
...
+        try {
+            InitializationService.initialize();
+            VerifySamlInitializer.init();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Test
+    public void allNullParametersShouldFailValidation() {
```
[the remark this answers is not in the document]
instantiator: 👍

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/ValidSamlId.java
instantiator: I'll do a little digging in case it's possible, but it didn't seem trivial to adjust the message from the annotation. The solution there might be to compose a validation from several smaller ones. It's quite boilerplate-y, though, which is why I'm hesitating...

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriStringValidator.java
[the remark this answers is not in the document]
JohnWattsGDS: Got you.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/ValidSamlId.java
instantiator: It's possible 👍

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
```diff
@@ -0,0 +1,36 @@
...
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (StringUtils.isBlank(value)) { return true; } // @NotNull should detect nulls
+
+        try {
+            String decoded = Base64.decodeAsString(value);
+            if (StringUtils.isBlank(decoded)) {
+                return false;
+            }
+
+            Object saml = new SamlParser().parseSamlString(decoded);
```
tomrosier: Can we be more specific than just object?

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public static final int MinLength = 64;
+    public static final int MaxLength = 64000;
+
+    @Override
+    public void initialize(ValidBase64Saml constraint) { /* intentionally blank */ }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (StringUtils.isBlank(value)) { return true; } // @NotNull should detect nulls
```
tomrosier: The comment at the end is redundant. Is the expected value of value to be a base64 encoded string?
Even an empty base64 string will have a length of 4.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
```diff
@@ -0,0 +1,36 @@
...
+public class Base64SamlValidator implements ConstraintValidator<ValidBase64Saml, String> {
+
+    public static final int MinLength = 64;
+    public static final int MaxLength = 64000;
+
+    @Override
+    public void initialize(ValidBase64Saml constraint) { /* intentionally blank */ }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
```
tomrosier: Could probably be more descriptive than 'value' maybe 'encodedSaml'

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: The value could be null or empty or whitespace, so the comment is to indicate that we're accepting empty values in this validator, with the expectation that a developer will use @NotNull if they don't allow nulls.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
```diff
@@ -0,0 +1,36 @@
...
+        try {
+            String decoded = Base64.decodeAsString(value);
+            if (StringUtils.isBlank(decoded)) {
+                return false;
+            }
+
+            Object saml = new SamlParser().parseSamlString(decoded);
+            return saml instanceof ResponseImpl;
+
+        } catch (Exception e) {
```
tomrosier: Can we be less generic with the exception?

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64XmlValidator.java
```diff
@@ -0,0 +1,36 @@
...
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (StringUtils.isBlank(value)) { return true; } // @NotNull should detect nulls
+
+        String xml = Base64.decodeAsString(value);
+        if (StringUtils.isEmpty(xml)) { return false; }
+
+        try {
+            InputStream stream = new ByteArrayInputStream(xml.getBytes());
+            DocumentBuilder parser = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+            Document document = parser.parse(stream);
```
tomrosier: Is there a check for checking if the document is empty?

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: Sadly, no. If we were going to do that, we'd write a validator for each possible SAML message returned - but we're not validating the content of the SAML here, we're validating that the parameter passed to us contains _some_ SAML.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriStringValidator.java
```diff
@@ -0,0 +1,33 @@
...
+
+        try {
+            URI uri = URI.create(value);
+            if (uri.getScheme().equalsIgnoreCase("http") ||
+                uri.getScheme().equalsIgnoreCase("https")) {
+                return true;
+            } else {
```
tomrosier: wrapping in else is redundant

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriStringValidator.java
```diff
@@ -0,0 +1,33 @@
...
+import java.net.URI;
+
+public class DestinationUriStringValidator implements ConstraintValidator<ValidDestinationUriString, String> {
+
+    @Override
+    public void initialize(ValidDestinationUriString constraint) { /* intentionally blank */ }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (StringUtils.isBlank(value)) { return true; } // @NotNull should detect nulls
```
tomrosier: Feels like it should be returning false.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: I see your point - although to me, it's just a String (or nothing) until we've validated it. Happy to change it if it's helpful to clarify.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriValidator.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.ida.notification.validations;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.net.URI;
+
+public class DestinationUriValidator implements ConstraintValidator<ValidDestinationUri, URI> {
```
tomrosier: How does this class differ from `DestinationUriStringValidator`?

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: The intention here was to catch _anything_ that goes wrong when trying to parse the String. A generic Exception seemed to capture that intention more clearly. Happy to change it if you think it's better to use the specific subclasses.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/LevelOfAssuranceValidator.java
```diff
@@ -0,0 +1,33 @@
...
+    public void initialize(ValidLOA constraint) {
+        this.acceptVerifyLOA = constraint.acceptVerifyLOA();
+        this.acceptEidasLOA = constraint.acceptEidasLOA();
+    }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (StringUtils.isBlank(value)) { return true; } // @NotNull should detect nulls
+
+        return
```
tomrosier: Loose the new line at the end of this line. ```
return (acceptVerifyLOA && EnumUtils.isValidEnum(VspLevelOfAssurance.class, value)) ||
           (acceptEidasLOA && EnumUtils.isValidEnum(EidasLoaEnum.class, value));
```

File: proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/PEMCertificateValidator.java [hunk 14]
```diff
@@ -0,0 +1,46 @@
...
+            context.disableDefaultConstraintViolation();
+            context.buildConstraintViolationWithTemplate("This certificate is not yet valid.").addConstraintViolation();
+            return false;
+
+        } catch (CertificateExpiredException e) {
+            context.disableDefaultConstraintViolation();
+            context.buildConstraintViolationWithTemplate("This certificate has expired.").addConstraintViolation();
+            return false;
+
+        } catch (Exception e) {
```
tomrosier: Don't think we should just catch all exceptions

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/PEMCertificateValidator.java
```diff
@@ -0,0 +1,46 @@
...
+            return false;
+        }
```
tomrosier: You can have `return false;` here rather than in each catch clause

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: On further inspection, `parseSamlString` is a generic method, and I'd need to provide it with a concrete type. I could possibly use `ResponseImpl`. I'm not sure it adds much beyond a bit of casting. What do you think?

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64XmlValidator.java
instantiator: This is checking to see if the content is an XML document. All XML documents must have a root element or they'll throw during parsing. We could add an additional validation for "and cannot be a valid XML document with only the root element" but it seemed a bit excessive.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriStringValidator.java
instantiator: It felt clearer to me. I'm happy to remove the wrap.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriStringValidator.java
instantiator: Ah, because my intent (and comment) hasn't been clear enough. This validator shouldn't respond false on a nullable parameter that is null. @NotNull is for preventing null values.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/DestinationUriValidator.java
instantiator: It operates on a URI instead of a String. Apparently we're using both.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/LevelOfAssuranceValidator.java
instantiator: 👍

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/PEMCertificateValidator.java
instantiator: Let's chat 1:1 - justification for similar decision above applies here too, but happy to resolve in conversation.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/PEMCertificateValidator.java
instantiator: 👍 can fix

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: (done)

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/PEMCertificateValidator.java
instantiator: 👎 just confirmed this with IntelliJ - java exceptions don't fall through, so I do need the return false in each exception clause.

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: (done for all)

proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/Base64SamlValidator.java
instantiator: None of the code here actually throws an exception. As we're inside a validator, my feeling is that we should capture any runtime exceptions, and return false for anything unexpected that means it can't resolve the Base64 String into SAML. I'm going to leave it in, and comment my decision.

PR #236, opened 2019-06-28
EID-1487 Send Istio headers with ProxyNodeJsonClient requests

JohnWattsGDS:
Create an IstioHeaderStorage class that uses ThreadLocal to store Istio headers.  Use a request filter to capture the Istio headers on the way in.  Update ProxyNodeJsonClient so that it has access to an instance of IstioHeaderStorage for retrieving the current request’s Istio headers to pass on.  Append Istio headers to response on the way out and clear the headers out of the thread local.

With added whitespace because failed merges.


CharlesIC:
Already merged in another PR

PR #237, opened 2019-07-01
Revert EID-1487 - Make the Proxy Node pipeline great again

idrop:
Reverts alphagov/verify-proxy-node#232.

We merged with devs outside the trusted set and will merge again with same approvers when all are trusted.


CharlesIC:
Lewis is now trusted, no need to revert John's changes

PR #240, opened 2019-07-02
Firebreak: Finesse the Proxy Node pipeline

CharlesIC:
A number of small improvements to the pipeline:
  1) Host metadata statically from apps

  - The changes in this PR will re-publish the existing metadata on the new URLs. If this works, the entity IDs will be changed in a separate PR
  - If this works we can remove a bunch of stuff from the pipeline
  - Added a configurable `MetadataPublishingBundle` that allows to publish a specified resource at a specified URL
  - The metadata resources will be mounted by K8 in the apps' resource directories in their pods

  2) Redirect to different error page per environment
  2) Make Stub Connector prettier and add favicon
  2) Make all Dockerfiles consistent and `wget` less noisy
  2) Make failed package builds not increase the release number until they go green (ripped from #230) 

I based 1) above on the following sources of information:
[Kubernetes docs - hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath)
[VMC config](https://github.com/alphagov/verify-metadata-controller/blob/f2862119d8b79e9fcbfc937c137cb039afda2747/pkg/controller/metadata/metadata_controller.go)
[Istio VirtualService reference](https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/#VirtualService)
[Article on routing in VirtualServices](https://rinormaloku.com/istio-practice-routing-virtualservices/)
@chrisfarms


chart/templates/gateway-virtual-service.yaml
```diff
@@ -13,6 +13,14 @@ spec:
   gateways:
   - gsp-gsp-cluster.gsp-system
   http:
+    - match:
+      - uri:
+          exact: {{ include "gateway.host" . }}/ServiceMetadata
```
idrop: Does the metadata url need to match the entityId of that metadata?
i.e. the entityId of https://test-metadata.london.verify.govsvc.uk/metadata.xml is "https://test-metadata.london.verify.govsvc.uk/metadata.xml". See also https://eidas.minez.nl/EidasNodeC/ConnectorMetadata
I'm not certain its required, but does seem to be a convention.

ci/build/release-pipeline.yaml
```diff
@@ -379,6 +379,7 @@ spec:
             RELEASE_NAME: test
             RELEASE_NAMESPACE: ((namespace-deployer.namespace))
             CLUSTER_PRIVATE_KEY: ((cluster.privateKey))
+            ERROR_PAGE_URL: https://www.integration.signin.service.gov.uk/proxy-node-error
```
idrop: 👍

idrop (dismissed):

chart/templates/gateway-virtual-service.yaml
CharlesIC: Yes it does but we can change that later - this will publish the existing metadata at the new URL in addition to the old one. If this works, we'll change the entity IDs and remove old metadata

idrop: Ok, makes sense

chart/templates/gateway-deployment.yaml
```diff
@@ -55,3 +67,7 @@ spec:
           value: http://{{ .Release.Name }}-vsp/
         - name: PROXY_NODE_ERROR_PAGE_URI
           value: {{ .Values.gateway.errorPageURL }}
+        - name: PROXY_NODE_METADATA_RESOURCE_FILE_PATH
```
chrisfarms: does `PROXY_NODE_METADATA_RESOURCE_FILE_PATH ` and `PROXY_NODE_METADATA_PUBLISH_PATH ` really need to be configurable via the environment?.... looks like they could just be hardcoded or could they have decent defaults so that the amount of config could be reduced.

chart/templates/stub-connector-deployment.yaml
```diff
@@ -53,6 +64,10 @@ spec:
           value: "80"
         - name: CONNECTOR_NODE_BASE_URL
           value: https://{{ include "connector.host" . }}
+        - name: CONNECTOR_METADATA_RESOURCE_FILE_PATH
```
chrisfarms: Same comment as per `PROXY_NODE_METADATA_RESOURCE_FILE_PATH`

chart/templates/gateway-deployment.yaml
CharlesIC: Yeah maybe they could - I'm still trying to figure if this will work and will refine this further

tommrr (dismissed):
LGTM

idrop (dismissed):

idrop (dismissed):

idrop (approved):

PR #242, opened 2019-07-03
Proxy Node Integration deployment using stub-connector

idrop:
**!!! Merge alphagov/verify-hub-federation-config#404 first !!!**

At present, there can only be a single connector per Proxy Node instance.

This PR creates a deployment of Proxy Node to Integration where the stub-connector is faking being the EU Member state. This will make integration testing easier for us as there will be no dependencies to other Member States.

To help review this PR, [here's a diff](https://github.com/alphagov/verify-proxy-node/commit/3ee29161ce7651e855ca27d98cabf893a50d6b14) between the NL integration deployment and this deployment.

A subsequent PR will tear down the build Proxy Node instance after the build pipeline is complete.

  - Restart production deployments at night to minimise disruption to users
  - Use one consistent entityID for the proxy node instead of different entity IDs for the Hub and the EUMS Connector Node


ci/integration/deploy-test.yaml
```diff
@@ -0,0 +1,135 @@
+---
+apiVersion: concourse.k8s.io/v1beta1
+kind: Pipeline
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: deploy
```
idrop: There [already is a pipeline](https://ci.london.verify.govsvc.uk/teams/proxy-node-integration/pipelines/deploy) whose metadata is `deploy`.  I'm unsure whether this new pipeline name needs to be distinct, or will slot into the existing `deploy` pipeline.

ci/integration/deploy-test.yaml
```diff
@@ -0,0 +1,135 @@
...
+              gpg --import key
+              gpg --export-secret-keys > ~/.gnupg/pubring.gpg
+              KEY_ID="$(gpg --list-secret-keys --with-colons  | awk -F: '/uid:/ {print $10}' | head -n1)"
+              echo "verifying package"
+              helm verify ./release/*.tgz
+              echo "rendering chart with release name '${RELEASE_NAME}' and namespace '${RELEASE_NAMESPACE}'..."
+              helm template \
+                --name "${RELEASE_NAME}" \
+                --namespace "${RELEASE_NAMESPACE}" \
+                --set "global.cluster.name=${CLUSTER_NAME}" \
```
CharlesIC: We need to also set the errorPageURL to integration URL. Just add it here like in my PR

ci/integration/deploy-test.yaml
CharlesIC: You should add this job to the `deploy-pipeline.yml` file. It's going to be just another job in the same pipeline.
But maybe Concourse is smart enough to combine things from different files 🤷‍♂
@chrisfarms thoughts?

ci/integration/deploy-test.yaml
```diff
@@ -0,0 +1,135 @@
...
+kind: Pipeline
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: deploy
+spec:
+  exposed: true
+  pipelineString: |
+
+    github_source: &github_source
```
CharlesIC: If you add this to the existing file then you don't have to define things like `&github_source` or `&task_toolbox` as they're already defined. I suggest moving this job to the existing pipeline file rather than having it in a separate file.

CharlesIC (changes requested):
Looks good but this should go into the `integration/deploy-pipeline.yml` file

ci/integration/deploy-test.yaml
idrop: Good spot 👀

CharlesIC (dismissed):
Looks legit 👍

ci/prod/deploy-pipeline.yaml
```diff
@@ -38,17 +38,17 @@ spec:
...
 
-    - name: daily
+    - name: nightly
       type: time
       icon: update
       source:
-        interval: 12h
-        start: 8:00 AM
-        stop: 8:00 PM
+        interval: 1h
```
philandstuff: it's a super minor thing, but I don't think this `interval:` line is required: https://github.com/concourse/time-resource#trigger-once-within-time-range

That said, happy to leave it be

philandstuff (commented):
I've left a mostly-useless comment. I'm going to defer to @samcrang as I'm wfh today and he has more recent understanding of this pipeline

ci/prod/deploy-pipeline.yaml
CharlesIC: I'm happy to leave it. It's the same mechanism we're using for the daily build and it works fine so no reason to change it

Wynndow (approved):

JohnWattsGDS (approved):

PR #243, opened 2019-07-03
👋Bye Bye local PKI

idrop:
We no longer build a local PKI, so this directory is confusing (and terrifying).

We'll either replace this with gsp-local or a sandbox dev instance.


tommrr (approved):

CharlesIC:
I'm thinking of writing a script to spin up the PN locally - either by using pure Docker or some simple K8 cluster. Could any of the stuff be useful for that? Or should we just use the sandbox dev?

CharlesIC:
We'll keep it for now as we may need to re-use it

PR #244, opened 2019-07-05
Fix to permit NL cert to pass size validation

instantiator:
- NL cert comes in at just over 2070 bytes.
- Increase upper bound on cert size to 4096, previous bound was 2048.


proxy-node-shared/src/main/java/uk/gov/ida/notification/validations/PEMCertificateValidator.java
```diff
@@ -12,9 +12,10 @@
 
 public class PEMCertificateValidator implements ConstraintValidator<ValidPEM, String> {
 
-    // self-signed test cert is ~900-1000 chars
+    // Self-signed test cert is ~900-1000 chars.
+    // NL cert comes in at just over 2070 chars.
```
CharlesIC: Tom's right - we don't put comments in the code except in very rare circumstances where it makes sense. We can get rid of these

PR #258, opened 2019-07-12
BAU: Serve Metadata from the local filesystem

CharlesIC:
This wasn't working previously as the java app resources live in the JAR and we can't mount the metadata there.
Instead, the app will serve from an arbitrary location in the container.

Also small improvements:
  - Limit ingress, ingress and istio logging to journey related resources
  - Add a simple resource to handle `favicon.ico` requests
  - Log absolute path in exception mappers
  - Fix Stub Connector favicon and logo size issue


proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceTest.java
```diff
@@ -118,7 +119,7 @@ private void verifyHappyPath() {
         verify(sessionStore).createOrUpdateSession(eq(sessionId), any(GatewaySessionData.class));
         verify(session).getId();
 
-        verify(appender).doAppend(captorILoggingEvent.capture());
+        verify(appender, times(1)).doAppend(captorILoggingEvent.capture());
```
idrop: Isn't times(1) implied?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/HubResponseResourceTest.java
```diff
@@ -47,13 +47,13 @@ public void testsHappyPath() {
             "eidas_request_id_in_session",
             "issuer",
             "connector_public_cert_in_session",
-            "http://conector.node"
+            "http://connector.node"
```
idrop: 🙈

proxy-node-gateway/src/test/java/uk/gov/ida/notification/resources/EidasAuthnRequestResourceTest.java
CharlesIC: Huh you're right, it is

idrop (approved):
Looks good, left one pedantic query on a test

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/HealthCheckAppRuleTests.java
```diff
@@ -1,29 +1,43 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
 
+@RunWith(MockitoJUnitRunner.class)
 public class HealthCheckAppRuleTests extends GatewayAppRuleTestBase {
 
     private static final String METADATA_PUBLISH_PATH = "/proxy-node-md-publish-path";
-    private static final String METADATA_RESOURCE_PATH = "metadata/test-proxy-node-metadata.xml";
+    private static final String METADATA_File_PATH =
+            HealthCheckAppRuleTests.class.getClassLoader().getResource("metadata/test-proxy-node-metadata.xml").getPath();
```
instantiator: Unusual capitalisation. Not really important to change, though!

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/AssetsResource.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.ida.notification.resources;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Response;
+
+@Path("/")
+public class AssetsResource {
```
instantiator: This would have made a superb #good-first-issue card for a dev new to DropWizard ;)

instantiator (approved):
Looks good. This was helpful to learn about the @Provider annotation for the ingress/egress filter, and about filters themselves. It isn't 100% clear to me why we're doing ingress/egress logging - but as this PR limits it to fewer things, that's a question for another time 😉 

The Assets/favicon Resource is a good example of a ticket that we could hand to a new dev to learn about DropWizard... 🆕🌟 We should think of more of these for when we bring brand new devs onto the team 👍

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/HealthCheckAppRuleTests.java
CharlesIC: Ugh I'll fix it in the next PR

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/AssetsResource.java
CharlesIC: Actually, it would probably be more elegant to handle those requests with a filter rather than have a whole new resource so feel free to take it on as a "learning Dropwizard" side project and make it better :)

PR #263, opened 2019-07-16
Limit scope of hub egress rules to current namespace only.

blairboy362:
Fix an issue with overlapping rules getting in each other's way.


blairboy362:
This change was live-edited in the Verify cluster to check it worked; and the tests succeeded.

PR #267, opened 2019-07-17
BAU: Use the new metadata published by the apps

CharlesIC:
- Changed entity IDs to match the new metadata endpoints
  - Rename secrets in K8s cluster to be more clear by better describing what purpose they serve
  - Add a separate metadata signing cert for Stub Connector


chart/templates/metadata-ingress-virtual-service.yaml
```diff
@@ -16,6 +16,6 @@ spec:
   http:
     - route:
       - destination:
-          host: {{ .Release.Name }}-proxy-metadata
+          host: {{ .Release.Name }}-proxy-node-metadata
```
blairboy362: This is pointing at the service that is created by the VMC when processing a `Metadata` resource, which might not be what we want. Potentially lengthy discussion to have around this that I'm not going to attempt via github comments.

CharlesIC: Actually we don't need this anymore - that's the thing I was talking about this morning. Let's discuss tomorrow.

chart/templates/_helpers.tpl
```diff
@@ -5,7 +5,7 @@
 
 {{- define "connector.host" -}}
```
blairboy362: Thinking about it...this block can be simplified further. We no longer need the `if...else`; just the `printf` inside the `if` part.

chart/templates/_helpers.tpl
```diff
@@ -5,7 +5,7 @@
...
 {{- if .Values.stubConnector.enabled -}}
```
blairboy362:
```suggestion
```

chart/templates/_helpers.tpl
```diff
@@ -5,7 +5,7 @@
...
-{{- printf "%s-%s.%s" .Release.Name "connector" (required "global.cluster.domain is required" .Values.global.cluster.domain) | trimSuffix "-" -}}
+{{- printf "%s" (include "connector.metadata.host" .) | trimSuffix "-" -}}
 {{- else -}}
```
blairboy362:
```suggestion
```

chart/templates/_helpers.tpl
```diff
@@ -5,7 +5,7 @@
...
 {{- printf "%s" .Values.connector.host -}}
```
blairboy362:
```suggestion
```

chart/templates/_helpers.tpl
```diff
@@ -5,7 +5,7 @@
...
 {{- end -}}
```
blairboy362:
```suggestion
```

blairboy362:
You may also be able to remove "metadata.host" from the `_helpers.tpl` (confirm it with a search, but I hope it is unused now).

blairboy362 (approved):

idrop (approved):

PR #272, opened 2019-07-20
BAU: Fix Stub Connector entity ID mismatch causing invalid signature errors in ESP

CharlesIC:
Added a config value to the Stub Connector to ensure it uses the same entity ID in its SAML messages as is advertised in its metadata

Also make chart in the build pipeline a proper pre-release so Concourse won't publish the new release before it passes the tests, as per the [Concourse GitHub Release Resource documentation](https://github.com/concourse/github-release-resource).

Bump OpenSAML and Dropwizard versions.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -2,16 +2,8 @@
 version: v1.13.5
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
 ignore:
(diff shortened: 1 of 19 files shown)


Todai88 (approved):
LGTM 🎉 🇮🇷

samcrang (approved):

idrop (approved):

PR #275, opened 2019-07-23
BAU: Publish Metadata signing certs from K8s Metadata secret

CharlesIC:
There are no more GSP admins so we need to have a way to extract the metadata certs to send to countries.

The `metadataCACerts` file is mounted from the cluster. It contains a base64 encoded string containing certs in PEM format. `MetadataCertsPublishingResource` reads the file, decodes it and then extracts the certs and publishes them on the specified URL.

This is what the certs page looks like
Screen Shot 2019-07-23 at 15 54 40


blairboy362:
Not sure I fully understand the problem you're trying to solve...so some rambling thoughts instead.

In my mind these CA certs should be sent to the various countries "out of band" in order to establish trust. So you don't necessarily want these published publicly. This is not a problem on its own; just...odd.

If this is just for our benefit, why don't we solve the problem properly by updating the metadata controller to put these values into a config map instead of a secret?

chart/templates/gateway-deployment.yaml
```diff
@@ -67,3 +69,8 @@ spec:
           value: http://{{ .Release.Name }}-vsp/
         - name: PROXY_NODE_ERROR_PAGE_URI
           value: {{ .Values.gateway.errorPageURL }}
+        - name : PROXY_NODE_METADATA_SIGNING_CERT
+          valueFrom:
+            secretKeyRef:
```
idrop: I ❤️  that you can do this

CharlesIC:
@blairboy362 We don't have access to config maps.
Also, these certs are published on an eIDAS portal so anyone on that portal can see them. They're public certs - they don't have  to be secret so no issue with publishing them like this. We could make the endpoint only accessible from the office network is this is a concern - but this can be done at a later time. What we need right is an easy way to access those certs.

blairboy362:
@CharlesIC If you can't access configmaps that is a bug. According to the permissions matrix you should have get/list/watch on configmaps.

I don't care about securing their access...they're public certs. That wasn't my point.

idrop (approved):

JohnWattsGDS (approved):

stub-connector/src/test/java/uk/gov/ida/notification/apprule/StubConnectorAssetsAppRuleTests.java
```diff
@@ -33,6 +33,13 @@ public void shouldServeMetadata() throws IOException, URISyntaxException {
         assertThat(metadata).isEqualTo(expectedMetadata);
     }
 
+    @Test
+    public void shouldPublishMetadataSigningCertificates() throws URISyntaxException {
+        final Response response = stubConnectorAppRule.target(METADATA_CERTS_PUBLISH_PATH).request().get();
+
+        assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
```
instantiator: Should this test also confirm that the content of the response is sensible?

instantiator (approved):

stub-connector/src/test/java/uk/gov/ida/notification/apprule/StubConnectorAssetsAppRuleTests.java
CharlesIC: We're testing it in the Gateway but sure - I'll add more assertions to this in another PR

PR #280, opened 2019-07-26
BAU: Misc fixes to Denmark pipeline and certs publisher

CharlesIC:
- Add missing DK deployment config
- Remove `HUB_FQDN` as `www.integration.signin.service.gov.uk` is the default value in `values.yml`
- Give deployments different unique names
- Filter certs on sanitised values
- Make cert naming consistent
- Rename metadata resolver to correct name


blairboy362 (approved):
The non-Java parts!

tommrr (approved):
LGTM

idrop (approved):

PR #284, opened 2019-07-26
EID-1566: Publish metadata trust chain certs

CharlesIC:
- Create certs from PEM strings instead of base64 to reduce the amount of sanitising necessary before decoding the certs
  - Mount the metadata signing cert as a file and filter it out from the published certs
  - Cache the rendered certs view in memory
  - Update the certs page styles to render the certs in the PEM format with a monospace font
  - Add resource rule tests for the certs publishing resource

This change will produce a page with metadata trust chain certs that looks like this:
Screen Shot 2019-07-26 at 20 02 41


tommrr (approved):
LGTM

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/metadata/MetadataCertsPublishingResource.java
```diff
@@ -67,51 +77,36 @@ public MetadataSigningCertsView getMetadataSigningCerts() {
...
-                .map(c -> c.replace("\n", ""))
-                .map(String::trim)
-                .filter(c -> !c.equals(metadataSigningCertBase64.replace("\n", "").trim()))
+                .map(c -> BEGIN_LINE + c + END_LINE)
                 .map(this::createCert)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
                 .collect(Collectors.toList());
 
+        readFile(metadataSigningCertFilePath).flatMap(this::createCert).ifPresentOrElse(
```
Wynndow: A comment here might be good to explain what the stream is doing. It looks like it's removing the metadata signing cert from the list of concatenated pem certs?

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/metadata/MetadataCertsPublishingResource.java
```diff
@@ -24,40 +25,49 @@
...
-        this.metadataSigningCertBase64 = metadataSigningCertBase64;
+        this.metadataSigningCertFilePath = metadataSigningCertFilePath;
         this.metadataCACertsFilePath = metadataCACertsFilePath;
         this.metadataPublishPath = metadataPublishPath;
     }
 
     @GET
     public MetadataSigningCertsView getMetadataSigningCerts() {
+        if (metadataSigningCertsView == null) {
+            metadataSigningCertsView = generateMetadataSigningCertsView();
```
Wynndow: If the view is cached in this way it tightly couples the rotating of the certs with a release of the app. Is this what we want?

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/metadata/MetadataCertsPublishingResource.java
CharlesIC: Yep that's what it does 👍

Wynndow: A comment would help for people reading it in future.

CharlesIC: Sure 👍

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/metadata/MetadataCertsPublishingResource.java
CharlesIC: I think we always have to release the apps to rotate the certs - so I think this is okay

CharlesIC: But we can revisit this in the future when we implement the rotation process

Wynndow (approved):

PR #292, opened 2019-07-31
Fix VSP secret to have extra layer of base64 encoding of the keys

Krenair:


Krenair:
For the record, when the sealed secrets controller unsealed this it correctly assumed `type: Opaque`, and also the change of ciphertext for the environment seems to be irrelevant - the value has not changed and this has been checked.

PR #293, opened 2019-08-02
EID-1553: Set correct intermediate metadata cert validity period

CharlesIC:
Intermediate metadata certs in prod should have a validity of 5 years, the same as our current certs.
SAML signing certs are rotated every 3 months.

### DIFF
--- a/chart/templates/metadata.yaml
+++ b/chart/templates/metadata.yaml
@@ -23,7 +23,7 @@ spec:
   samlSigningCertRequest:
     countryCode: GB
     commonName: Verify Proxy Node SAML Signing
-    expiryMonths: 2
+    expiryMonths: 3
     organization: Cabinet Office
     organizationUnit: GDS
     location: London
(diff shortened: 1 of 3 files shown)


michaelabenyohai (dismissed):

kerrr (dismissed):

ci/integration/metadata-ca-cert-request-integration.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   countryCode: GB
   commonName: Proxy Node Integration Metadata CA
-  expiryMonths: 120
```
Wynndow: The PR comment mentions that prod CA should have an expiry of 5 years. This updates integration but looks like it leaves prod at 10?

CharlesIC: Let me double check that!

PR #302, opened 2019-08-09
[DO NOT MERGE] BAU: stop pipeline stalling if `test` fails

bjgill:
Currently, if `test` fails, then the release will remain in the pre-release state. This means that it will not get picked up by the new run of the pipeline. Thus, the next run of package will just republish the old release and thus test will never trigger again.

We discovered this bug in the DCS pipeline, which is very similar to the proxy node's. See https://github.com/alphagov/doc-checking/commit/c32445c95c055545228013904f2ca25bafafd769 for the equivalent change to the DCS pipeline.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -98,6 +98,14 @@ spec:
         pre_release: true
         release: false
         
+    - name: last-release
+      type: github-release
+      icon: file-document
+      source:
+        <<: *github_source
+        pre_release: true
+        release: true
+        
     - name: release
       type: github-release
       icon: tag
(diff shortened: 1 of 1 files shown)


CharlesIC:
I've already solved this problem in our pipeline. We have the `chart` which is the `last-release` in your PR and a `release` which is the release being prepared. I will review this to see if we've missed anything.

bjgill:
I think the problem is that the 'generate_release-version' task uses 'release' to get the version number of the previous release. Happy to discuss in person if that'd be easier...

CharlesIC:
We've got a different solution to this so this change is not needed. Closing PR.

PR #303, opened 2019-08-09
🍪 EID-1599 - Add a session cookie to Proxy Node gateway

idrop:
What
----
Add a cookie to gateway which will contain a signed and encrypted [JSON Web Token](https://en.wikipedia.org/wiki/JSON_Web_Token) (JWT) set in a cookie.
The JWT will contain the following data:
* `issuer`, the entity id of EU Member State
* `eidasRequestId`, sent by EU Member State
* `eidasRelayState,` sent by EU Member State
* `hubRequestId` generated by the VSP service

This cookie will in time replace the existing session storage in redis, in order to make the gateway stateless and help provide Zero Downtime Deployments (ZDD).

The cookie will be set as `HttpOnly`, `Secure`, have a lifetime of 90 minutes, and be limited to the current domain.

A subsequent PR will remove gateway-redis and http sessions on the gateway.

**Additionally**, data captured by Member State Connector Metadata will be saved in-memory in each gateway instance, keyed to the entity id of the connector. This data is:
* `encryptionCertificate` - the public key used to encrypt eidas response for a  Member State
* `eidasDestination` - the eidas response endpoint

This data structure is described by `EntityMetadata.java`.

The method for signing and encrypting a JWT is taken from <https://connect2id.com/products/nimbus-jose-jwt/examples/signed-and-encrypted-jwt>.

 How to review
-------------
The PR is broken down into commits to making reviewing more palatable.

 The most important commit to review is [critical changes to review](https://github.com/alphagov/verify-proxy-node/commit/99bbd1419b694083453c77be52f61336b739b0f6). 

Thereafter [config changes](https://github.com/alphagov/verify-proxy-node/commit/c1537aed8680bfc4dda4183e207a86e3fbc52e86) is important to understand.


proxy-node-gateway/src/main/java/uk/gov/ida/notification/RemoveSessionCookieFilter.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain chain) throws IOException, ServletException {
+        chain.doFilter(servletRequest, servletResponse);
+        HttpServletRequest request = (HttpServletRequest) servletRequest;
+        Cookie[] cookies = request.getCookies();
+        if (cookies != null) {
+            Arrays.stream(cookies)
+                    .filter(c -> COOKIE_X_PN_SESSION.equals(c.getName()))
+                    .findFirst()
```
Wynndow: It's possible for there to be more than one cookie with the same name. I'm not sure if we'd ever get into that scenario but it might be worth invalidating anything with the right name and not just the first in the stream.

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/EntityMetadata.java
```diff
@@ -0,0 +1,30 @@
...
+        this.data = new HashMap<>();
+    }
+
+    public synchronized void setData(String issuer, Key key, String value) {
+        Map<Key, String> entry = data.getOrDefault(issuer, new HashMap());
+        entry.put(key, value);
+        data.put(issuer, entry);
+    }
+
+    public synchronized String getValue(String issuer, Key key) {
```
Wynndow: I think this might leave us with a few issues that having a redis solved.
- If we have more than one instance of the gateway running, if the gateway receives a hub response and it's processed by an instance that hasn't yet seen an authn request from that issuer, this will throw an IllegalStateException. Assuming high traffic from issuers this should be fairly rare, but it's possible.
- If we release the gateway at any point this map will be lost, and hub responses will throw an IllegalStateException (if no new authn requests have been seen from that issuer by the gateway instance processing the request)

Could we store the encryption cert and the destination in the cookie with the other attributes? I'm not sure what the storage limit is...

proxy-node-gateway/src/main/java/uk/gov/ida/notification/resources/HubResponseResource.java
```diff
@@ -72,4 +85,41 @@ public View hubResponse(
             sessionData.getEidasRelayState()
         );
     }
+
+    private void checkClaimsMatchGatewaySessionData(HttpServletRequest request, GatewaySessionData sessionData) {
+        try {
+            Map<String, Object> claims = sessionCookieService.getData(request.getCookies());
+            if (!sessionDataMatches(sessionData, claims)) {
+                ProxyNodeLogger.warning("sessionData and sessionDataFromCookie are not equal");
```
Wynndow: Might be helpful to log out the diff between the two?

rhowe-gds: Only if the data they contain isn't/can't be sensitive

Wynndow: It's not, just some request id's, a relay state and an issuer.

PR #313, opened 2019-08-11
BAU: Add local startup configuration

CharlesIC:
This adds the ability to easily run the Proxy Node apps locally with minimal set-up using Docker. There are certain limitations to this at the moment, described in the instructions.

Recommend reviewing the `local-startup/running-proxy-node-locally.md` file first.

  - Provide Dockerfiles for local running
  - Provide `docker.env` file to set up apps for local running
  - Add `docker-compose.yml` file to run Proxy Node apps easily
  - Add script to start up Proxy Node
  - Add local running instructions
  - Update gradle to latest version
  - Add metadata files - these have been manually signed with `xmlsectool` and cannot be changed without re-signing
  - Add metadata truststores - these have been generated manually with `keytool` and need to be re-generated if metadata gets re-signed with a different cert chain
  - Add local PKI (keys and certs symlinked from `verify-dev-pki`)
  - Use locally built artefacts in Docker images by default 
  - Pull VSP image from Proxy Node registry rather than building it from scratch

To test a local journey run `./docker-startup.sh --build` and open `http://localhost:6610/` in your browser. See README for instructions on how to deal with the Compliance Tool.


local-startup/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN chmod +x /tini
+
+ARG component
+COPY --from=build /app/${component}/build/install/${component} .
+
+ENV CONFIG_FILE config.yml
+ENV COMPONENT $component
+
+ENTRYPOINT ["/tini", "--"]
+CMD "bin/$COMPONENT"
```
rhowe-gds: recommend `["bin/$COMPONENT"]` to avoid wrapping it in `sh -c`

local-startup/jdk11-jre-image/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN chmod +x /tini
+
+ARG component
+COPY --from=build /app/${component}/build/install/${component} .
+
+ENV CONFIG_FILE config.yml
+ENV COMPONENT $component
+
+ENTRYPOINT ["/tini", "--"]
+CMD "bin/$COMPONENT"
```
rhowe-gds: same `CMD` style recommendation as above

local-startup/running-proxy-node-locally.md
```diff
@@ -0,0 +1,81 @@
...
+$ docker-compose build SERVICE
+
+# For example
+$ docker-compose build translator
+```
+
+**Build images using locally built artifacts**
+```shell script
+# Tell Dockerfile to copy artifacts instead of building them
+$ export USE_LOCAL_BUILD=true
```
rhowe-gds: If `USE_LOCAL_BUILD` only applies to `docker-compose`, you can run it as

`USE_LOCAL_BUILD=true docker-compose ...` which will avoid polluting your shell environment

proxy-node-vsp-config/verify-service-provider.yml
```diff
@@ -28,11 +28,11 @@ hashingEntityId: ${HASHING_ENTITY_ID:-}
 verifyHubConfiguration:
   environment: ${VERIFY_ENVIRONMENT:-}
 
-samlSigningKey: ${SAML_SIGNING_KEY:-}
+samlSigningKey: ${VSP_SAML_SIGNING_KEY:-}
```
rhowe-gds: Will the service work if this is blank? If not, why even have the fallback? Have dropwizard throw an error due to a missing interpolation token instead.

docker-compose.yml
```diff
@@ -0,0 +1,95 @@
...
+      dockerfile: ../verify-proxy-node/proxy-node-vsp-config/Dockerfile
+      args:
+        RUN_TESTS: ${RUN_TESTS:-false}
+        VERIFY_USE_PUBLIC_BINARIES: ${VERIFY_USE_PUBLIC_BINARIES:-false}
+    env_file: ./local-startup/docker.env
+    environment:
+      JAVA_OPTS: "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:50302"
+    command: ["development", "-u", "http://localhost:6600/SAML2/SSO/Response/POST"]
+    ports:
+      - "50300:50300"
```
tomrosier: Is it worth switching theses to 6*** port just to be inline with the other ones.

local-startup/running-proxy-node-locally.md
```diff
@@ -0,0 +1,81 @@
+# Running proxy node locally
+
+This directory contains resources to run a Proxy Node deployment locally with Docker. This is a first attempt and while a whole journey can be successfully completed, there are certain limitations:
+
+- The Translator (and all other components) is configured to use a file credential configuration, which means it uses a set of pk8 files on the disk to sign and encrypt messages. Hence, the HSM functionality is not tested. The next step would be to investigate whether we can run a [SoftHSM](https://www.opendnssec.org/softhsm/) instance locally to simulate a real HSM for the Translator.
```
tomrosier: We should discuss the SoftHSM stuff, as it wasn't appropriate when we tried to use it before as its implementation is vastly different from the AWS HSM implementation.

startup-docker.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Starting Gateway, Translator, Stub Connector, VSP..."
+docker-compose up -d stub-connector proxy-node-gateway translator verify-service-provider
+
+# ESP won't start without Connector metadata
+printf "Waiting for Stub Connector"
+until $(curl --output /dev/null --silent --head --fail http://localhost:6610/ConnectorMetadata); do
+  printf '.'
+  sleep 2
+done
+echo
```
tomrosier: Is this redundant?

tommrr (commented):
LGTM, will approve after Russells changes have been applied

rhowe-gds:
> LGTM, will approve after Russells changes have been applied

Don't feel you need to hold off on my comments - they're merely suggestions and I don't think there'll be any real change in behaviour

docker-compose.yml
CharlesIC: Not really - VSP isn't our app. It's fine if it has a different port number.

local-startup/running-proxy-node-locally.md
CharlesIC: Yeah we might not be able to use it. I've not looked into it in much detail but we should definitely look into it.

startup-docker.sh
CharlesIC: Nope - because above we use `printf` so there's no newline at the end

local-startup/Dockerfile
CharlesIC: We need variable substitution, though. [Docker docs](https://docs.docker.com/engine/reference/builder/#cmd) say "if you want shell processing then either use the shell form or execute a shell directly". So it's got to be the _exec_ form or this:
`CMD ["/bin/sh", "-c", "\"bin/${COMPONENT}\""]`

JohnWattsGDS (approved):
This is probably fine, if it doesn't work, I'll find out when I try to run it locally.

CharlesIC:
Simply run `./startup-docker.sh --build` - it's magic! 🎩✨

idrop (approved):
🚀

PR #316, opened 2019-08-16
BAU: Refactor tests

CharlesIC:
This is a fairly big PR to refactor the tests and make them faster and more robust. I recommend reviewing by PRs as all changes have been split into digestible chunks by component. The highlights are:

**Make things static where possible**
We used to initialise a lot of things before each test. This causes high memory and CPU usage, especially when spinning up DropwizardAppRules (test apps). Making everything initialise only once reduces the tests runtime by half (!) and prevents the Concourse workers from dying from lack resources

 **Don't share AppRules and test clients between tests**
Due to how JUnit and Dropwizard initialise resources, there are multiple issues when those resources are shared between different test classes. Each test class now has its own set of resources to avoid concurrency and interdependency issues

**Apps that use metadata need to wait for it before running tests**
Due to how Dropwizard and JUnit initiate resources, test metadata client may not be ready to serve metadata when the tests start. Added a custom method to wait for it to become available before initialising resources dependent on it


tommrr (approved):

.travis.yml
```diff
@@ -21,7 +21,7 @@ install:
   - npm install -g snyk@1.146.0
 
 script:
-  - ./gradlew --no-daemon test
+  - ./gradlew --no-daemon --parallel test
```
JohnWattsGDS: Are we convinced that parallel and static aren't going to clash?

JohnWattsGDS (approved):

proxy-node-shared/src/test/java/uk/gov/ida/notification/saml/EidasResponseBuilderTest.java
```diff
@@ -28,7 +28,7 @@
         try {
             InitializationService.initialize();
         } catch (InitializationException e) {
-            throw new IllegalStateException("Could not initialize opensaml in test", e);
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
CharlesIC: It doesn't really matter here, we just want to know if the tests have run and passed.

.travis.yml
CharlesIC: Yep it's fine. Concurrency issues have been fixed now.

PR #325, opened 2019-08-27
removes redundent config from chart

chrisfarms:
## What

Removes redundent VirtualServices and DestinationRules from proxy-node chart.

## Why

communication with the cloudhsm is now restricted with a GlobalNetworkPolicy that requires pods to have the `talksToHsm: true` label instead of Istio VirtualServices, DesintaitonRules and Gateways
per namespace which means we can remove this config and simplify things a bit.

## :warning: Before Merge

Depends on https://github.com/alphagov/gsp/pull/414 being rolled out first!


blairboy362:
https://github.com/alphagov/gsp/pull/414 has been rolled out

PR #326, opened 2019-09-02
BAU: Set correct intermediate metadata cert validity period

CharlesIC:
**Set correct intermediate metadata cert validity period**
Intermediate metadata certs in prod should have a validity of 5 years, the same as our current certs. SAML signing certs are rotated every 3 months.

**Remove filtering out of leaf metadata signing cert from the published certs**
This is no longer necessary as the VMC has been modified to not include the leaf certs in the CA certs secret. Remove caching of certs view as CA certs can now be mounted dynamically by the K8s cluster.


idrop (dismissed):

idrop (approved):
"Appease Snyk" 👍

chart/templates/metadata.yaml
```diff
@@ -19,11 +19,12 @@ spec:
     contactGivenName: GOV.UK Verify
     contactSurname: Support
     contactEmail: idasupport@digital.cabinet-office.gov.uk
+    validityDays: 14
   enabled: true
   samlSigningCertRequest:
     countryCode: GB
     commonName: Verify Proxy Node SAML Signing
-    expiryMonths: 2
+    expiryMonths: 3
```
Wynndow: Shouldn't this be 9 months? The plan is they expire after 9 months but we rotate after 3, right?

CharlesIC: I guess but if we plan to rotate them every 3 months then it's effectively the same. We can change this to 9 months if needed when we implement the cert rotation in VMC

ci/build/build-pipeline.yaml
```diff
@@ -262,6 +262,7 @@ spec:
           trigger: true
 
       - task: run-tests
+        attempts: 2
```
JohnWattsGDS: Out of curiosity; why are we now attempting the tests twice?

Wynndow (approved):

ci/build/build-pipeline.yaml
CharlesIC: Cos they randomly fail sometimes 🤷‍♂

JohnWattsGDS: As good a reason as any!

paroxp: 😬

PR #328, opened 2019-09-03
🇳🇱 Update NL int Connector Node truststore

CharlesIC:
Update Dutch integration truststore to contain trust chain rather than metadata signing leaf cert

Dutch Int CN metadata:
https://acc-eidas.minez.nl/EidasNodeC/ConnectorMetadata

To verify certs in the truststore:
```
$ echo "<trust-store-base64>" | base64 -D > nl-int.truststore
$ keytool -keystore nl-int.truststore -storepass <default-store-pass-from-pn-repo> -list -v
```

To get the password search for `metadataSigningTruststorePassword`  in `values.yaml`


Wynndow (approved):
Have looked at the trust store and can see that the metadata signing cert used to sign the metadata here https://acc-eidas.minez.nl/EidasNodeC/ConnectorMetadata, is signed with the top most certificate in the trust store with CN `QuoVadis PKIoverheid Organisatie Services CA - G3` .

idrop (approved):

PR #343, opened 2019-09-19
EID-1682: Add in pipeline to murder the proxy node on que

tomrosier:
⚠️ ⚠️ ⚠️ This creates a pipeline that can nuke the proxynode. ⚠️ ⚠️ ⚠️


tomrosier:
Its set as paused when its deployed.

PR #344, opened 2019-09-26
Invalidate a http session on a successful response from Hub

idrop:
Invalidate the http session after a successful egress from Gateway to the Government Service.


CharlesIC:
To do tomorrow:
  - Invalidate session also for the two error mappers
  - Or ideally invalidate only in place for all exit paths 
  - Test that session gets invalidated in all three scenarios

PR #348, opened 2019-10-07
Trigger package after VSP build step

Krenair:
Since you can't run any image other than the latest from Harbor, if we build a new image we need to ensure it gets deployed.


CharlesIC:
This is not necessary - `build-base-images` will trigger `build` through `cloudhsm-jce-image` which will then trigger `package` and `release`

CharlesIC:
If you merge this, each build of `build-base-images` will result in two releases instead of one.

CharlesIC:
pn-build

CharlesIC:
Superseded by #349

PR #351, opened 2019-10-09
Present pipelines as proper YAML instead of a string

CharlesIC:
The `config` option is now supported instead of `pipelineString: |`. This allows the pipelines to be proper YAML which means we get validation, syntax highlighting and less errors due to stray whitespaces.


Krenair (approved):

Wynndow (approved):
Awesome.

idrop (approved):
What the world's been waiting for

PR #353, opened 2019-10-10
BAU: Add cert checking script

tomrosier:
It checks metadata for countries to see if its valid.


CharlesIC:
We'll handle metadata validation as part of the monitoring work. Closing.

PR #354, opened 2019-10-11
🇲🇹 Add Proxy Node deployment for Malta in integration

CharlesIC:


JohnWattsGDS (approved):

Wynndow:
Where did the trust-anchor come from? I'm struggling to see it on the CEF page and can't see it in the zendesk ticket.

CharlesIC:
@Wynndow From the metadata itself - they publish a different cert on CEF - I'm assuming they haven't updated CEF. But in any case the published cert is a leaf metadata singing cert so instead I just took the leaf cert from the current published metadata.

Wynndow:
I don't think we want to trust something that we just grabbed from the metadata. We should be getting it from them directly, or from their CEF page.

CharlesIC:
You're right - but for integration this is okay

Wynndow (approved):
Yep, fair enough

PR #355, opened 2019-10-14
[DO NOT MERGE] BAU Remove CZ integration deployment

Wynndow:
The deployment is currently failing due to the CZ connector-node
metadata endpoint returning a 404. This causes the esp container to
crash loop. We're getting constant alerts in the alerts channel which
are noisy and may make genuine issues harder to spot.

Once the endpoint has been fixed, this commit can be reverted.


CharlesIC:
Can we instead delete the deployment in the cluster form command line? And pause the deployment pipeline for CZ in Concourse so it doesn't get reapplied to the cluster. This way we don't have to do two PRs to remove and add it again. Also - just merging this may not delete the actual deployment in the cluster anyway so manual deletion for the time being is the best course of action.

Wynndow:
@CharlesIC I've tried deleting the deployment manually, but as it's in the Verify cluster I don't have permissions to do it.

PR #357, opened 2019-10-22
EID-1542 Apply GOV.UK style and wording to the JavaScript disabled 'Continue' page

JohnWattsGDS:
Copy and paste the required styles from the Verify version of these pages.

Remove the 'submitText' parameter from the template, the view that uses the template and anything else affected by the demise of the 'submitText' parameter.

Actually using the full stylesheets would require all sorts of extra work for a page that nobody is going to see unless they've disabled JavaScript so we can use inline styles here because rules need to be applied sensibly.

### DIFF
--- a/proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/SamlResponseExceptionMapperResourceRuleTest.java
+++ b/proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/SamlResponseExceptionMapperResourceRuleTest.java
@@ -37,12 +37,14 @@ public class SamlResponseExceptionMapperResourceRuleTest {
 
     @Before
     public void before() {
-        when(sessionStore.getSession(any(String.class))).thenReturn(
-                new GatewaySessionData("HubRequestId", "EidasRequestId", "EidasDestination",
-                        "EidasConnectorPublicKey", "EidasRelayState"));
-
-        when(samlFormViewBuilder.buildResponse("EidasDestination", (String) null, "Continue", "EidasRelayState"))
-                .thenReturn(new SamlFormView("postUrl", "samlMessageType", "encodedSamlMessage", "submitText"));
+        when(sessionStore.getSession(any(String.class))).thenReturn(new GatewaySessionData("HubRequestId","EidasRequestId","EidasDestination","EidasConnectorPublicKey","EidasRelayState"));
+        String nullString = null;
+        when(samlFormViewBuilder.buildResponse("EidasDestination", nullString, "EidasRelayState"))
+                .thenReturn(
+                new SamlFormView("postUrl",
+                                 "samlMessageType",
+                                 "encodedSamlMessage",
+                                 "submitText"));
     }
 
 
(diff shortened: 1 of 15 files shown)


proxy-node-shared/src/main/resources/uk/gov/ida/notification/views/saml-form.mustache
```diff
@@ -15,40 +15,81 @@
...
-        cursor: pointer;
-      }
-      .submitBtn:hover, .submitBtn:focus {
-        background-color: #00692f;
-      }
+        body {
+            padding-top: 2em;
+            padding-left: 2em;
+            background-color: white;
+            font-family: "nta", Arial, sans-serif;
```
jakubmiarka: are we importing the `nta` font?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/SamlResponseExceptionMapperResourceRuleTest.java
```diff
@@ -37,12 +37,14 @@
...
     public void before() {
-        when(sessionStore.getSession(any(String.class))).thenReturn(
-                new GatewaySessionData("HubRequestId", "EidasRequestId", "EidasDestination",
-                        "EidasConnectorPublicKey", "EidasRelayState"));
-
-        when(samlFormViewBuilder.buildResponse("EidasDestination", (String) null, "Continue", "EidasRelayState"))
-                .thenReturn(new SamlFormView("postUrl", "samlMessageType", "encodedSamlMessage", "submitText"));
+        when(sessionStore.getSession(any(String.class))).thenReturn(new GatewaySessionData("HubRequestId","EidasRequestId","EidasDestination","EidasConnectorPublicKey","EidasRelayState"));
+        String nullString = null;
+        when(samlFormViewBuilder.buildResponse("EidasDestination", nullString, "EidasRelayState"))
```
galund: I assume this was done to shut up codacy or something, but I feel like it should have been a separate PR as it doesn't seem to have anything to do with this one?

proxy-node-shared/src/main/resources/uk/gov/ida/notification/views/saml-form.mustache
```diff
@@ -15,40 +15,81 @@
...
-        font-size: 1em;
-        line-height: 1.25;
-        border: none;
-        box-shadow: 0 2px 0 #003618;
-        cursor: pointer;
-      }
-      .submitBtn:hover, .submitBtn:focus {
-        background-color: #00692f;
-      }
+        body {
```
galund: Wouldn't it be easier just to link or import some CSS file that's already deployed somewhere else in Verify?

proxy-node-shared/src/main/resources/uk/gov/ida/notification/views/saml-form.mustache
```diff
@@ -15,40 +15,81 @@
...
     </style>
 </head>
 
 <body onload="init()">
 <form action={{postUrl}} method=post name=saml-form>
-    <h1>Continue to next step</h1>
-    <p>Because Javascript is not enabled on your browser, you must press the continue button</p>
+    <h1 class="heading-large">Continue to next step</h1>
+    <div class="application-notice info-notice">
+        <p>Because Javascript is not enabled on your browser, you must press the continue button</p>
```
galund: What about Welsh :-p

proxy-node-shared/src/main/resources/uk/gov/ida/notification/views/saml-form.mustache
CharlesIC: @jakubmiarka are those stylesheets deployed somewhere?

proxy-node-shared/src/main/resources/uk/gov/ida/notification/views/saml-form.mustache
CharlesIC: Lol I actually don't think that page has a Welsh version even in the Hub 😱

CharlesIC: I stand corrected.... it does!
Screen Shot 2019-10-23 at 13 22 02

proxy-node-shared/src/main/resources/uk/gov/ida/notification/views/saml-form.mustache
jakubmiarka: they are but please don't hot link them.. they are versioned so the path is not guaranteed to be static...

galund: I take back what I said about 'easier'. 'More correct', maybe, but eh 😂

PR #360, opened 2019-10-27
Brewfile: Alphabetize, and `minikube` is no longer a Cask

issyl0:
**I am no longer a trusted dev since I removed myself after leaving Autom8, if that makes any difference here.**


Wynndow:
Isn't it the case that all commits to the repo need to be made my trusted devs? Or is having two approvals from trusted devs enough?

CharlesIC:
I don't know - if it fails we'll just wait for the next valid PR 🤷‍♂

issyl0:
You can pull this branch and rebase it and edit my commits and amend them with your signatures, if you like, but that feels like effort.

PR #364, opened 2019-11-04
EID-1687 Improve handling of missing first name attribute in Translator

idrop:
https://govukverify.atlassian.net/browse/EID-1687

There have been instances of the Proxy Node throwing “Error whilst handling Hub response. No verified current first name present”.

An attribute can be provided to a Member State if it is verified and valid for the present time, or if it is valid for the present time.

Currently an attribute is only provided in the former case.

1. Relax logic in the Attribute class to allow
    * the from date to be null,
    * allow no from and to date to be valid for the present time
2. Change translator logic to accept attributes in this order:
    * current and verified
    * If no attributes satisfying a), look for current (without
requiring verified)
3. Log when b) happens so we can track how many times this occurs
4. Log when no current with verified or unverified so we can track which scenarios we have to deal with

The attribute data to be logged is:
- 	the attribute type: first name, last name etc.
- 	is verified
- 	is validFrom not null
- 	is validTo not null
The attribute value will not be logged.

Note the use of the `var` keyword.

### DIFF
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attribute.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attribute.java
@@ -1,5 +1,6 @@
 package uk.gov.ida.notification.contracts.verifyserviceprovider;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.joda.time.DateTime;
@@ -54,13 +55,23 @@ public class Attribute<T> {
         return to;
     }
 
-
-    public boolean isValid() {
+    public boolean isVerifiedAndCurrent() {
         return this.verified && this.isCurrent() && this.value != null;
     }
 
     public boolean isCurrent() {
-        return (this.from != null && this.from.isBeforeNow()) &&
+        return (this.from == null || this.from.isBeforeNow()) &&
                 (this.to == null || this.to.isAfterNow());
+    }
+
+    @Override
+    @JsonIgnore
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("Attribute{");
+        sb.append("verified=").append(verified);
+        sb.append(", has from=").append(from != null);
+        sb.append(", has to=").append(to != null);
+        sb.append('}');
+        return sb.toString();
     }
 }
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attribute.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attribute.java
@@ -55,10 +55,6 @@ public class Attribute<T> {
         return to;
     }
 
-    public boolean isVerifiedAndCurrent() {
-        return this.verified && this.isCurrent() && this.value != null;
-    }
-
     public boolean isCurrent() {
         return (this.from == null || this.from.isBeforeNow()) &&
                 (this.to == null || this.to.isAfterNow());
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
@@ -106,16 +106,20 @@ public class Attributes {
 
         public List<Attribute<T>> getValidAttributes() {
 
-            var valid = getAllAttributes().stream()
-                    .filter(Attribute::isVerifiedAndCurrent)
+            var current = getAllAttributes().stream()
+                    .filter(Attribute::isCurrent)
                     .collect(Collectors.toList());
-            if (valid.isEmpty()) {
+
+            var verifiedAndCurrent = current.stream()
+                    .filter(Attribute::isVerified)
+                    .collect(Collectors.toList());
+
+            if (verifiedAndCurrent.isEmpty()) {
                 ProxyNodeLogger.info("No verified and current attributes: " + createAttributesMessage());
-                valid = getAllAttributes().stream()
-                        .filter(Attribute::isCurrent)
-                        .collect(Collectors.toList());
+                return current;
+            } else {
+                return verifiedAndCurrent;
             }
-            return valid;
         }
 
         public List<Attribute<T>> getAllAttributes() {
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
@@ -1,14 +1,16 @@
 package uk.gov.ida.notification.contracts.verifyserviceprovider;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.joda.time.DateTime;
 import org.joda.time.chrono.ISOChronology;
 import org.joda.time.format.ISODateTimeFormat;
+import uk.gov.ida.notification.shared.logging.ProxyNodeLogger;
 
 import javax.validation.constraints.NotNull;
-import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 public class Attributes {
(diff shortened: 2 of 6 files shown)


proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
```diff
@@ -53,40 +55,78 @@ public Attributes(
...
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+
+        public List<Attribute<T>> getValidAttributes() {
+
+            var valid = getAllAttributes().stream()
```
instantiator: 🎆

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
```diff
@@ -53,40 +55,78 @@ public Attributes(
...
-                .map(Attribute::getValue)
-                .filter(s -> !s.isEmpty())
-                .collect(Collectors.joining(" "));
+    @JsonIgnore
+    public AttributesList<Address> getAddresses() {
+        return new AttributesList(datesOfBirth, "dateOfBirth");
     }
 
     // Prints date in EIDAS format YYYY-MM-dd
     public static String getFormattedDate(DateTime date) {
```
instantiator: Should we rename this method `formatAsEidasDate`?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -119,39 +115,38 @@ private static String getMappedStatusCode(VspScenario vspScenario) {
...
+        if (validDatesOfBirth.isEmpty()) {
+            throw new HubResponseTranslationException("No verified current date of birth present: " + datesOfBirth.createAttributesMessage());
+        }
+        return validDatesOfBirth.stream()
                 .map(Attribute::getValue)
                 .reduce(BinaryOperator.maxBy(DateTimeComparator.getDateOnlyInstance()))
-                .map(Attributes::getFormattedDate)
-                .orElseThrow(() -> new HubResponseTranslationException("No verified current date of birth present"));
+                .map(Attributes::getFormattedDate).get();
     }
```
instantiator: Whilst this method does exactly what it says on the tin (fetches the date of birth that's the latest, from the current and valid subset of dates of birth), I'm not sure this is the most correct thing to do.

There should only ever be one _current_ entry for date of birth. If an IDP responds with two current dates of birth, it suggests that the IDP believes both can be simultaneously true. Fetching the one that's the latest date of birth seems less correct than rejecting the response as nonsensical. I might add:

```
if (validDatesOfBirth.size() > 1) {
    throw new HubResponseTranslationException("The IDP has provided more than one current date of birth.");
}
```

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
```diff
@@ -53,40 +55,78 @@ public Attributes(
...
+        AttributesList(List<Attribute<T>> attributes, String type) {
+            this.attributes = Optional.ofNullable(attributes).orElse(Collections.emptyList());
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+
+        public List<Attribute<T>> getValidAttributes() {
```
instantiator: `valid` here meaning `verified` and `current`, I wonder if it's clearer to refactor this to `getVerifiedAndCurrentAttributes`. I found `isVerifiedAndCurrent` to be clear about what it's doing.

instantiator (dismissed):
Looks good. Pretty much only cosmetic suggestions, and a conversation to have around how an IDP could possibly believe two dates of birth to be _current_.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
idrop: Let's debate this further in the office

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
```diff
@@ -53,40 +55,78 @@ public Attributes(
...
-
-    public static String combineAttributeValues(List<Attribute<String>> attributes) {
-        return attributes.stream()
-                .filter(Objects::nonNull)
-                .map(Attribute::getValue)
-                .filter(s -> !s.isEmpty())
-                .collect(Collectors.joining(" "));
+    @JsonIgnore
+    public AttributesList<Address> getAddresses() {
+        return new AttributesList(datesOfBirth, "dateOfBirth");
```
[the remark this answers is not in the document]
idrop: ugh, good catch, updated

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
```diff
@@ -53,40 +55,82 @@ public Attributes(
...
+            var verifiedAndCurrent = current.stream()
+                    .filter(Attribute::isVerified)
+                    .collect(Collectors.toList());
+
+            if (verifiedAndCurrent.isEmpty()) {
+                ProxyNodeLogger.info("No verified and current attributes: " + createAttributesMessage());
+                return current;
+            } else {
+                return verifiedAndCurrent;
+            }
```
CharlesIC: No need for the `else` part here since the `if` part ends in `return`

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attribute.java
```diff
@@ -54,13 +55,19 @@ public DateTime getTo() {
...
     }
 
-
-    public boolean isValid() {
-        return this.verified && this.isCurrent() && this.value != null;
-    }
-
     public boolean isCurrent() {
-        return (this.from != null && this.from.isBeforeNow()) &&
+        return (this.from == null || this.from.isBeforeNow()) &&
```
galund: This fixes the second A/C in https://govukverify.atlassian.net/browse/EID-1750 so we can update that ticket once this lands 👍

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -119,39 +115,38 @@ private static String getMappedStatusCode(VspScenario vspScenario) {
...
 
     private static String getLatestValidDateOfBirth(HubResponseContainer hubResponseContainer) {
-        return getValidAttributes(hubResponseContainer.getAttributes().getDatesOfBirth())
-                .stream()
+        var datesOfBirth = hubResponseContainer.getAttributes().getDatesOfBirth();
+        var validDatesOfBirth = datesOfBirth.getValidAttributes();
+        if (validDatesOfBirth.isEmpty()) {
+            throw new HubResponseTranslationException("No verified current date of birth present: " + datesOfBirth.createAttributesMessage());
+        }
+        return validDatesOfBirth.stream()
```
galund: This stuff potentially more belongs in the other ticket i.e. https://govukverify.atlassian.net/browse/EID-1750 – doesn't seem to relate to this ticket? Seems like the behaviour is unchanged though so I guess it doesn't matter.

(Out of interest, do we just not like the `orElseThrowThing` or is there a particular reason for this change? –it was new to me anyway, I would have always written it the way you now have it :-) )

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
```diff
@@ -53,40 +55,82 @@ public Attributes(
...
+            var current = getAllAttributes().stream()
+                    .filter(Attribute::isCurrent)
+                    .collect(Collectors.toList());
+
+            var verifiedAndCurrent = current.stream()
+                    .filter(Attribute::isVerified)
+                    .collect(Collectors.toList());
+
+            if (verifiedAndCurrent.isEmpty()) {
+                ProxyNodeLogger.info("No verified and current attributes: " + createAttributesMessage());
```
galund: I'm not clear why we might log at INFO level unless we are particularly interested in this case? We don't, for example, much care if there are no middle names?

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
idrop: Updated after our review together, thanks!

CharlesIC (approved):

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
CharlesIC: This is just FYI for us so we can diagnose why we're seeing missing first name errors for NL. This is an "expected error" but just need more details when it happens.

galund: OK, maybe put it a TODO to say that we're going to remove it again then? (or move the logging to somehow be specific to the first name case that we're interested in?)

Are we keeping the ticket open until we've actually checked the logs for this, then?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
idrop: `validDatesOfBirth.isEmpty()` is a method on a list - not an Optional. If it were an Optional we could do something like orElseThrow

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
CharlesIC: We want to keep it forever - and it's not specific to just the first name. Basically, we want to know when there are missing attribute values so we can investigate and handle that case better. This is fine as it is IMO

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
idrop: We discussed this at dev huddle and decided we would continue the behaviour of sending one date of birth

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/Attributes.java
galund: I definitely disagree that we want to create INFO log entries for ever for expected cases like user not having middle names. That just seems wrong to me.

CharlesIC: We can always adjust the logging in the future if needed. For now, we want to know when it happens

galund (approved):
We need to decide whether to cherry-pick this across to the release branch.

PR #366, opened 2019-11-05
fix istio metrics to actually work

philandstuff:
These istio metrics were originally created in #310 and worked
provided you only had a single helm release per namespace.  However
because it didn't include appropriate `{{ .Release.Name }}` scoping in
the resource names, it completely blocked the ability to deploy
multiple instances per namespace.

That problem was fixed in #322, which changed the resource names to
include `{{ .Release.Name }}` as appropriate.  However it didn't
update references to those resource names, nor did it update the name
of the underlying prometheus metric.  This means that the istio
metrics from #310 aren't currently being recorded at all.

This commit fixes the resource references to actually point at each
other, and also adds the `{{ .Release.Name }}` to the prometheus
metric name.  Hopefully, once this is merged we'll be able to see
metrics like `verify_proxy_node_prod_nl_requests_by_path_total` for
each release of the proxy node in each namespace.

### DIFF
--- a/chart/templates/metrics.yaml
+++ b/chart/templates/metrics.yaml
@@ -39,9 +39,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   actions:
-  - handler: prometheusbypath
+  - handler: {{ .Release.Name }}-prometheus-by-path-handler
     instances:
-    - requestcountbypath.metric
+    - {{ .Release.Name }}-metrics-request-count-by-path.metric
   match: (context.protocol == "http") && (match((request.useragent | "-"), "kube-probe*") == false)
 ---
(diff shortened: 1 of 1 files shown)


CharlesIC:
`verify_proxy_node_prod_nl_requests_by_path_total` it will work for now but once we have a single instance per env will this still be okay?

philandstuff:
Sure! The actual name is

> `{{ .Release.Namespace | replace "-" "_" }}_{{ .Release.Name | replace "-" "_" }}_requests_by_path_total`

which is determined by the namespace name and the release name.  Once you have a single instance, the metric will be determined by whatever release name you choose for the multi country proxy node deployment.

PR #369, opened 2019-11-08
Delete unused sealed secrets

idrop:
Delete unused JKWSets that were going to be used on encrypted data in a secure cookie.
This design was abandoned - these sealed secrets are unused.

Also remove the thing that generates the JKWSets 😞


CharlesIC:
Much better :)
We shouldn't have to run a thing to generate a piece of config for the app. That reeks of generating truststores out of certs instead of the app generating them itself

PR #372, opened 2019-11-14
EID-1542: Make the no-js "Continue" button the same as in the Hub

CharlesIC:
The Hub button says "Continue", not "Submit".
Also clean up the unnecessary args as the button text is no longer customisable.

### DIFF
--- a/proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/SamlResponseExceptionMapperResourceRuleTest.java
+++ b/proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/SamlResponseExceptionMapperResourceRuleTest.java
@@ -44,9 +44,8 @@ public class SamlResponseExceptionMapperResourceRuleTest {
                 new SamlFormView("postUrl",
                                  "samlMessageType",
                                  "encodedSamlMessage",
-                                 "submitText"));
+                                 "relayState"));
     }
-
 
     @Test
     public void shouldMapSessionAlreadyExistsExceptionToExceptionToSamlErrorResponseMapper() {
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/SamlFormViewBuilder.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/SamlFormViewBuilder.java
@@ -10,13 +10,13 @@ import uk.gov.ida.notification.views.SamlFormView;
 public class SamlFormViewBuilder {
     private SamlObjectMarshaller marshaller = new SamlObjectMarshaller();
 
-    public SamlFormView buildRequest(String url, AuthnRequest authnRequest, String submitText, String relayState) {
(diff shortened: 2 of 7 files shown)


JohnWattsGDS (approved):
This should be ~submitted~ continued with.

tommrr (approved):

CharlesIC:
Your reviews don't count @tomrosier - it's a grey tick 😕

proxy-node-shared/src/main/java/uk/gov/ida/notification/SamlFormViewBuilder.java
```diff
@@ -10,13 +10,13 @@
...
-    public SamlFormView buildRequest(String url, AuthnRequest authnRequest, String submitText, String relayState) {
+    public SamlFormView buildRequest(String url, AuthnRequest authnRequest, String relayState) {
         String samlMessage = marshaller.transformToString(authnRequest);
         String encodedSamlMessage = Base64.encodeAsString(samlMessage);
-        return buildRequest(url, encodedSamlMessage, submitText, relayState);
+        return buildRequest(url, encodedSamlMessage, relayState);
     }
 
-    public SamlFormView buildRequest(String url, String encodedSamlMessage, String submitText, String relayState) {
+    public SamlFormView buildRequest(String url, String encodedSamlMessage, String relayState) {
```
instantiator: It looks like `submitText` wasn't used in this method anyway?

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/SamlResponseExceptionMapperResourceRuleTest.java
```diff
@@ -44,10 +44,9 @@ public void before() {
                 new SamlFormView("postUrl",
                                  "samlMessageType",
                                  "encodedSamlMessage",
-                                 "submitText"));
+                                 "relayState"));
```
instantiator: This is interesting - it looks like you've fixed a bug. (My understanding was that the `SamlFormView` constructor would have dropped from 5 params to 4 params (removing the `submitText`), but this went from 4 to 4...

instantiator (approved):

proxy-node-shared/src/main/java/uk/gov/ida/notification/SamlFormViewBuilder.java
CharlesIC: Yep - this is just cleaning things up

proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/SamlResponseExceptionMapperResourceRuleTest.java
CharlesIC: Yep - there were some changes made a while ago to the SAML form view and not everything was cleaned up properly then

PR #373, opened 2019-11-14
EID-1762 Move attribute hash logging logic to saml-lib

Wynndow:
**This saml-lib PR will need merging and the new building importing before Travis will be happy: https://github.com/alphagov/verify-saml-libs/pull/102**

We log eIDAS attribute hashes in the proxy-node as well as the hub.
These hashes are compared for security in Splunk.

We want to ensure the same logic is used so the hashes will be the same.

The proxy-node receives attributes from the VSP. The VSP extracts those
attributes from an assertion and serializes them before sending in a
response.

This commit removes the custom Assertion and Attribute objects the
proxy-node has and replaces them with the same classes as used in the
VSP (via saml-lib). This way they should de-serialize into the same
objects. It also imports the new logger that is defined in saml-lib.


proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/AttributesBuilder.java
```diff
@@ -1,59 +1,64 @@
...
-    private List<Attribute<String>> lastNames = new ArrayList<>(singletonList(createAttribute(LAST_NAME)));
-    private Attribute<String> gender = createAttribute(GENDER);
-    private List<Attribute<DateTime>> datesOfBirth = new ArrayList<>(singletonList(createAttribute(DATE_OF_BIRTH)));
-
-    public Attributes build() {
-        return new Attributes(
+    public static final LocalDate VALID_FROM = createDateTime(2001, 1, 1);
+    public static final LocalDate DATE_OF_BIRTH = createDateTime(1990, 1, 1);
+
+    private List<NonMatchingTransliterableAttribute> firstNames = new ArrayList<NonMatchingTransliterableAttribute>(singletonList(createNonMatchingTransliterableAttribute(FIRST_NAME)));
```
JohnWattsGDS: I think you can get away with dropping the new ArrayList<...>(...) and just use the singletonList(...) utility method which will give you a List of the right type anyway for this and the other instances.

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponse.java
```diff
@@ -1,10 +1,12 @@
 package uk.gov.ida.notification.contracts.verifyserviceprovider;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.ida.eidas.logging.TranslatedHubResponseInterface;
+import uk.gov.ida.verifyserviceprovider.dto.NonMatchingAttributes;
 
 import javax.validation.constraints.NotNull;
 
-public class TranslatedHubResponse {
+public class TranslatedHubResponse implements TranslatedHubResponseInterface {
```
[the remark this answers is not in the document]
Wynndow: I would've but the interfaces have been dropped!

proxy-node-shared/src/test/java/uk/gov/ida/notification/contracts/verifyserviceprovider/AttributesBuilder.java
Wynndow: Intellij was hiding it from me, will remove.

Wynndow:
Closing in favour of https://github.com/alphagov/verify-proxy-node/pull/381

PR #374, opened 2019-11-15
remove commit_verification_keys

philandstuff:
The latest version of the github resource doesn't need this (since
alphagov/gsp#378).  Instead the process is to add your GPG key to
GitHub and get GitHub to verify the signature; we only need to
validate the signature on the GitHub merge commit against the GitHub
web flow key.

Note that if we can get rid of commit_verification_keys from all our
pipelines, we can also merge alphagov/gds-trusted-developers#111 so
that we no longer have to faff with GPG keys in that repository when
people join teams that commit to GSP projects.

This also bumps the task-toolbox and github-resource references to use
the tag provided by GSP in the `concourse` secret.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -18,7 +18,6 @@ spec:
       access_token: ((github.api-token))
       approvers: ((trusted-developers.github-accounts))
       required_approval_count: 2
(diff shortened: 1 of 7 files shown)


CharlesIC:
Please note this will have to be cherry picked to the `release/single-country-legacy` branch

PR #381, opened 2019-11-18
EID-1762 Move attribute hash logging logic to saml-lib

Wynndow:
We log eIDAS attribute hashes in the proxy-node as well as the hub.
These hashes are compared for security in Splunk.
We want to ensure the same logic is used so the hashes will be the same.

The proxy-node receives attributes from the VSP. The VSP extracts those
attributes from an assertion and serializes them before sending in a
response.

This commit removes the custom Attribute class and replaces it with
Attribute classes used by the VSP (via saml-lib). It also makes the
Attributes class extend the NonMatchingAttributes class which is what
the VSP uses to serialize the attributes.
This way they should de-serialize into the same
objects. It also imports the new logger that is defined in saml-lib.


proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponse.java
```diff
@@ -50,6 +53,6 @@ public VspLevelOfAssurance getLevelOfAssurance() {
     }
 
     public Attributes getAttributes() {
-        return attributes;
+        return attributes == null ? null : fromNonMatchingAttributes(attributes);
```
CharlesIC: It feels like the non-null check should be moved to the `fromNonMatchingAttributes` method - and return null from there if `attributes` is null. Actually in that case `fromNonMatchingAttributes` should return an Optional as this is the proper way to deal with missing values

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -57,7 +58,14 @@ public Response hubResponse(@Valid HubResponseTranslatorRequest hubResponseTrans
                         )
                 );
 
-        logResponseAttributesHash(hubResponseTranslatorRequest, translatedHubResponse);
+
+        EidasAttributesLogger eidasAttributesLogger = new EidasAttributesLogger(EidasResponseAttributesHashLogger::instance, null);
```
CharlesIC: I'm not sure this is the best way to instantiate the logger - as in the caller shouldn't know or care that there are different methods of instantiating the logger depending on which method of hashing is desired (e.g. that the second arg has to be null in this case). This could be abstracted away further. But this is okay for now and we can refactor this later.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -135,18 +135,19 @@ private static String getCombinedSurnames(HubResponseContainer hubResponseContai
...
             throw new HubResponseTranslationException("No verified current date of birth present: " + datesOfBirth.createAttributesMessage());
         }
+
         return validDatesOfBirth.stream()
-                .map(Attribute::getValue)
-                .reduce(BinaryOperator.maxBy(DateTimeComparator.getDateOnlyInstance()))
-                .map(Attributes::getDateInEidasFormat).get();
+                .map(NonMatchingVerifiableAttribute::getValue)
+                .reduce(BinaryOperator.maxBy(Comparator.comparing(LocalDate::toEpochDay)))
+                .map(date -> date.toString()).get();
```
CharlesIC: Why was the `getDateInEidasFormat` logic removed? There's a specific date format for eIDAS that we need to comply with.

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponse.java
Wynndow: I did wonder about that. Will update to an Optional.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
Wynndow: Previously we were using a joda date object, now it's a java LocalDate. Calling `toString` on LocalDate gives us the correct format, which is nice!

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VspProxyTranslateResponseTest.java
```diff
@@ -109,7 +110,7 @@ public void shouldReturnTranslatedHubResponseAuthnFailed() {
         assertThat(VspScenario.AUTHENTICATION_FAILED).isEqualTo(response.getScenario());
         assertThat("123456").isEqualTo(response.getPid());
         assertThat(VspLevelOfAssurance.LEVEL_2).isEqualTo(response.getLevelOfAssurance());
-        assertThat(response.getAttributes()).isNull();
+        assertThat(response.getAttributes().orElse(null)).isNull();
```
CharlesIC: Should we be testing for `response.getAttributes().isEmpty()`?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
CharlesIC: OK fine 👍

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VspProxyTranslateResponseTest.java
Wynndow: I suppose we should, although I think it's functionally the same.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -115,18 +115,18 @@ private static String getMappedStatusCode(VspScenario vspScenario) {
     }
 
     private static String getCombineFirstAndMiddleNames(HubResponseContainer hubResponseContainer) {
-        var firstNames = hubResponseContainer.getAttributes().getFirstNames();
+        var firstNames = hubResponseContainer.getAttributes().getFirstNamesAttributesList();
```
CharlesIC: Wouldn't this cause a NPE since you're doing `this.attributes = translatedHubResponse.getAttributes().orElse(null)` in `HubResponseContainer`? I think the Optional should carry through to here and be handled safely at the point of usage. If the attributes are `Optional.empty` then in the `getTranslatedHubResponse` method above we can skip a bunch of stuff

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VspProxyTranslateResponseTest.java
CharlesIC: Yes - it's more clear to use the Optional's nicely-named methods

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/verifyserviceprovider/TranslatedHubResponse.java
```diff
@@ -49,7 +54,7 @@ public VspLevelOfAssurance getLevelOfAssurance() {
         return levelOfAssurance;
     }
 
-    public Attributes getAttributes() {
-        return attributes;
+    public Optional<Attributes> getAttributes() {
+        return fromNonMatchingAttributes(attributes);
```
CharlesIC: Would `attributes` ever be null though since we have the `@NotNull` flag above?

Wynndow: 🤔 hmm, that's a good point. The main reason I'm handling the null case is due to [this test](https://github.com/alphagov/verify-proxy-node/blob/eid-1762-consolidate-user-attribute-hash-logging-logic/proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/VspProxyTranslateResponseTest.java#L101) which suggests that if authentication failed, you'd get a response with the attributes as null...

Wynndow: I've checked and VSP will return `null` for attributes, as well as `pid`, and  `loa`. I'm not sure what good the @NotNull flag is doing. Does it even ever get validated?

Wynndow: It doesn't. I don't think the annotations are actually doing anything. I've removed them.

CharlesIC: huh interesting. Good job for handling the null case 👍

CharlesIC: Should we also make `getLevelOfAssurance` and `getPid` Optional then?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -48,19 +49,20 @@ Response getTranslatedHubResponse(HubResponseContainer hubResponseContainer) {
         final String pid = pidPrefix + hubResponseContainer.getPid();
 
         if (hubResponseContainer.getVspScenario().equals(VspScenario.IDENTITY_VERIFIED)) {
+            var attributes = getAttributesOrThrow(hubResponseContainer);
```
CharlesIC: I don't think we need a separate method here - fine to just inline the logic in `getAttributesOrThrow`

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -46,10 +47,17 @@ public HubResponseTranslator(
     Response getTranslatedHubResponse(HubResponseContainer hubResponseContainer) {
         final List<EidasAttributeBuilder> eidasAttributeBuilders = new ArrayList<>();
 
-        final String pid = pidPrefix + hubResponseContainer.getPid();
+        final String pid = hubResponseContainer.getPid()
+                .map(p -> pidPrefix + p)
+                .orElse(null);
```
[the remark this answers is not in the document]
Wynndow: We should, but what this does is fairly analogues to the existing behaviour. I'm not inclined to start changing the logic of how we translate responses too much as it feels like creep of this story.

I've noticed as part of this work that we have an issue in the translator as to how we handle non-successful responses (the statuses don't map properly from what the VSP will return) and I'm going to write up a ticket to address it. I suspect looking at how we handle those scenarios (where these attributes can be null) should be part of that ticket.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -92,8 +100,11 @@ Response getTranslatedHubResponse(HubResponseContainer hubResponseContainer) {
     }
 
     @SuppressWarnings("SwitchStatementWithTooFewBranches")
-    private static String getMappedLoa(VspLevelOfAssurance vspLoa) {
-        switch (vspLoa) {
+    private static String getMappedLoa(Optional<VspLevelOfAssurance> vspLoa) {
+        if (vspLoa.isEmpty()) {
+            return null;
+        }
```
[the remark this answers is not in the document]
Wynndow: Probably not, but again this is analogues to the existing behaviour. This needs addressing but I think should be in another ticket which I'll write.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
```diff
@@ -46,10 +47,17 @@ public HubResponseTranslator(
     Response getTranslatedHubResponse(HubResponseContainer hubResponseContainer) {
         final List<EidasAttributeBuilder> eidasAttributeBuilders = new ArrayList<>();
 
-        final String pid = pidPrefix + hubResponseContainer.getPid();
+        final String pid = hubResponseContainer.getPid()
+                .map(p -> pidPrefix + p)
+                .orElse(null);
 
         if (hubResponseContainer.getVspScenario().equals(VspScenario.IDENTITY_VERIFIED)) {
```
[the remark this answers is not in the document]
Wynndow: Yep, I agree, but again, I think this should be done in a separate ticket now, as it's not super related to combining has logging logic.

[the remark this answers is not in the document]
galund: I think this is more or less this ticket, but we de-prioritised it: https://govukverify.atlassian.net/browse/EID-1793

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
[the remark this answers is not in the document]
CharlesIC: OK that's a good point - agree it's a good idea to do it in a separate story 👍

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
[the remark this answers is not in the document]
CharlesIC: it's not just the pid but it's part of the validation story Chris will write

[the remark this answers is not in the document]
galund: Sure, you might want to capture it in the existing ticket though, otherwise that one will be redundant :-)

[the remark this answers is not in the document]
Wynndow: Said story: https://govukverify.atlassian.net/browse/EID-1849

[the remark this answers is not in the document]
Wynndow: Yeah, it looks like that ticket @galund linked is covered by the one I just wrote. I might abandon that one in favour of my new one which covers a few other bits if people are happy with that?

Wynndow:
Have updated the PR to be against the release branch. Once it's landed I'll merge it back into master via another PR

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseTranslator.java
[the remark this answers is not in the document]
CharlesIC: The new story links to EID-1793 and largely covers it so I think it's fine to abandon the old story. We will have a reference to it when we come back to this.

PR #387, opened 2019-11-19
EID-1822 empty commit to test legacy PN release process

galund:
- this should generate version v1.117 of the proxy node

https://govukverify.atlassian.net/browse/EID-1822


galund:
We inadvertently tested this anyway and release v1.116 got deployed already 🤷‍♂

PR #376, opened 2019-11-18
BAU: build with latest java 11 on amazonlinux

idrop:
Install `java-11-amazon-corretto` on the container against latest `amazonlinux`.

This fixes the `javax.net.ssl.SSLPeerUnverifiedException: peer not authenticated` error we were getting when connecting to https://plugins.gradle.org:443 using TLS 1.3.

A failed example build that this PR fixes is https://ci.london.sandbox.govsvc.uk/teams/proxy-node-dev/pipelines/build-release/jobs/build-proxy-node/builds/60

PR #389, opened 2019-11-19
EID-1822: Use the Amazon Corretto🍦image as the base image for apps that need Java

CharlesIC:
This image is provided by Amazon and has JDK 11 installed with some Java patches.
The logic to install Java has been removed as the new image comes with Java pre-installed.

This replaces #376

**Also see #390 for more Corretto🍦goodness**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,7 @@
 # Base AWS Java app image
 
-ARG JAVA_HOME=/usr/local/openjdk-11
-FROM openjdk:11-jre-slim AS jre
-
-FROM amazonlinux:2.0.20190508
+FROM amazoncorretto:11
 WORKDIR /app
-
-# Install Java
-ARG JAVA_HOME
-ENV JAVA_HOME $JAVA_HOME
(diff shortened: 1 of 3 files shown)


adityapahuja (approved):
I have checked amazoncorretto:11 image. It does have JAVA_HOME environment variable set. 🌟

instantiator (approved):

galund:
There are still some tasks using an 'openjdk' image. I wonder if some or all of them should be using corretto as well.

See https://github.com/alphagov/verify-proxy-node/blob/eid-1822-use-amazon-corretto-image/ci/build/build-pipeline.yaml#L51

`run-tests` should surely be using corretto as well (and I don't see why it isn't using the main `Dockerfile`?) – I'm less sure about whether the VSP should be running in corretto or not, but potentially it'd be more consistent?

CharlesIC:
@galund see #390

galund (approved):

galund (approved):

PR #396, opened 2019-11-25
EID-1822 build and run with utf8 file encoding

idrop:
### Why
[Integration tests are failing](https://ci.london.verify.govsvc.uk/teams/proxy-node-build/pipelines/build-release/jobs/build-proxy-node/builds/32) to correctly compare the string contents of a `favicon.ico` to expected output.

The Amazon Coretto Docker image is not setting file-encoding to utf-8, which presumably was set in openjdk images.

### What
* Set the file encoding to utf-8 in `GRADLE_OPTS` to allow the tests to pass
* Also set the file-encoding of the built applications to use utf-8

This fix has been tested in sandbox, and fixes the pipeline in that cluster.

-----------
This was a useful test failure as it surfaced an encoding issue that might cause issues in production.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -219,6 +219,7 @@ spec:
               - |
                 cd vsp-src
                 rm -f .dockerignore
+                export GRADLE_OPTS="-Dfile.encoding=utf-8"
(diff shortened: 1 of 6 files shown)


adityapahuja:
Corretto-8 is using UTF-8 but Corretto-11 is not. 

Source: https://github.com/corretto/corretto-8-docker/pull/24

PR #406, opened 2019-12-02
EID-1823 Add VSP keys as sealed secret to new build namespace

idrop:
Add VSP encryption and signing keys as sealed secrets to the Verify cluster, `verify-eidas-proxy-node-build` namespace.

This will enable our `build` Proxy Node VSP to sign SAML requests to the Hub and decrypt SAML responses from the Hub.

We need to add a new sealed secret to the Verify cluster as it is scoped to the new `verify-eidas-proxy-node-build` namespace, tracking the `master` branch.

Another PR will be raised for the yet to be created integration and production namespace.

Please see https://verify-team-manual.cloudapps.digital/documentation/support/eidas-proxy-node/connecting-vsp-to-hub.html for detail on how secrets are sealed.


CharlesIC:
🤞

.snyk
```diff
@@ -5,7 +5,7 @@ ignore:
   SNYK-JAVA-CAJULIUSDAVIES-30073:
     - '*':
         reason: Fix not available
-        expires: 2019-11-30T19:21:05.816Z
+        expires: 2019-12-15T00:00:00.000Z
```
Wynndow: Might be worth giving this a longer lifetime? A couple of months perhaps?

idrop: I'd rather have us address it more often. If its a deal-breaker I can bump it to a month or two

Wynndow: Fair do's. The original ignore was added 5 months ago, so I reckon we'll end up bumping quite a lot though.

CharlesIC: If fix is not available then I think it's fine to wait a few months. We don't want the builds to constantly break because of this

PR #408, opened 2019-12-03
EID-1848 Deploy countries configMap

Wynndow:
This PR should not be merged before this: https://github.com/alphagov/verify-eidas-config/pull/2

This adds tasks to the build pipeline to create a configMap for
consumption by the proxy-node's metadata service, containing the
configuration for the countries.

The config is defined in the `verify-eidas-config` repo, as well as the
tooling to generate the truststore required. The resulting config file
from running the tool is used to populate a configMap.

We use kubectl with the `--dry-run` flag to generate the yaml resource
definition. We do this as we also use the `--from-file` flag - it's not
possible to define a configMap with static kubeyaml that references a
file in this way.

We edit the output of kubectl with `yq` to add a couple of annotations.
These annotations are used by `kapp`. They allow `kapp` to version the
configMap. This is required so that when we deploy the metadata-service
it is aware of which specific version of the configMap it is using. This way,
when the configMap changes and is redeployed, `kapp` knows that
metadata-service needs to also be redeployed - to pick up the latest
version of the configMap. The docs for this versioning [can be found
here](https://github.com/k14s/kapp/blob/master/docs/faq.md#updating-deployments-when-configmap-changes)

The versioning is only necessary as the metadata service currently only
reads the config file (mounted from the configMap) on startup. If it
could be designed to automatically "hot-reload" from the file we could
skip this, as the mounted config file is automatically updated when the
configMap is updated.

The final yaml resource definition for the configMap is added to the
`manifests` directory, which is where the other manifests are placed by
the `helm template` command. This allows the existing call to `kapp` to
deploy the configMap along with the apps themselves.

When we have the new pipelines for deploying to integration and prod we can
copy the pattern used here to update configMaps used for int and prod.

Co-authored-by: John Watts <john.watts@digital.cabinet-office.gov.uk>

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -541,8 +541,7 @@ spec:
               - -euc
               - |
                 cd eidas-config/tools
-                ./gradlew run --args="--mode=config --environment=TEST"
-                mv config.yml countriesConfig.yml
+                ./gradlew run --args="generate-config --environment=test"
 
       - task: create-config-map
         config:
@@ -571,9 +570,9 @@ spec:
                 kubectl config use-context deployer
 
                 echo "generating configMap resource definition from file"
-                kubectl -n "${NAMESPACE}" create configmap test-countries-config --from-file=countriesConfig.yml -o yaml --dry-run \
+                kubectl -n "${NAMESPACE}" create configmap test-countries-config --from-file=countriesConfig.yaml -o yaml --dry-run \
                   | yq -y '.metadata += {"annotations": {"kapp.k14s.io/versioned": "", "kapp.k14s.io/num-versions":"5"}}' \
-                  | tee ../../manifests/countriesConfigMap.yml
+                  | tee ../../manifests/countriesConfigMap.yaml
 
       - task: extract-chart
         image: chart
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -536,12 +536,11 @@ spec:
           outputs:
             - name: eidas-config
           run:
-            path: /bin/sh
+            dir: eidas-config/tools
+            path: gradlew
             args:
-              - -euc
-              - |
-                cd eidas-config/tools
-                ./gradlew run --args="generate-config --environment=test"
+              - run
+              - --args="generate-config --environment=test"
 
       - task: create-config-map
         config:
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -92,6 +92,16 @@ spec:
         <<: *github_source
         paths: [proxy-node-vsp-config]
 
+    - name: eidas-config
+      type: git
+      icon: bugle
+      source:
+        uri: https://github.com/alphagov/verify-eidas-config.git
+        branch: master
+        paths: [test]
+        username: "re-autom8-ci"
+        password: ((github.api-token))
+
     - name: cloudhsm-config
       type: github
       icon: folder-key-network-outline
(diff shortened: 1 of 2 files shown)


ci/build/build-pipeline.yaml
```diff
@@ -92,6 +92,16 @@ spec:
         <<: *github_source
         paths: [proxy-node-vsp-config]
 
+    - name: eidas-config
+      type: git
+      icon: bugle
+      source:
+        uri: https://github.com/alphagov/verify-eidas-config.git
+        branch: master
+        paths: [test]
```
idrop: I ❤️this scoping to the test directory. For the deploy pipeline, we might need to separate integration and production changes a bit differently..

Wynndow: My thinking was we'd have two different pipelines for integration and prod, each with their own resource definitions? If not, and they're both in the same pipeline, we could just define two resources for the different config dirs?

JohnWattsGDS (dismissed):

ci/build/build-pipeline.yaml
```diff
@@ -514,6 +524,55 @@ spec:
...
+          inputs:
+            - name: eidas-config
+          outputs:
+            - name: eidas-config
+          run:
+            path: /bin/sh
+            args:
+              - -euc
+              - |
+                cd eidas-config/tools
```
galund: If the only reason for using `sh` is so that we can change the current working directory, then it may be feasible and less code to use the `dir` option to `run` instead. https://concourse-ci.org/tasks.html#task-run-dir

ci/sandbox/build-pipeline.yaml
```diff
@@ -502,6 +512,55 @@ spec:
...
+
+                echo "configuring kubectl"
+                echo "${KUBERNETES_SERVICE_ACCOUNT}" | jq -r .["ca.crt"] > ca.crt
+                kubectl config set-cluster self --server=https://kubernetes.default --certificate-authority=ca.crt
+                kubectl config set-credentials deployer --token "${KUBERNETES_TOKEN}"
+                kubectl config set-context deployer --user deployer --cluster self
+                kubectl config use-context deployer
+
+                echo "generating configMap resource definition from file"
+                kubectl -n "${NAMESPACE}" create configmap test-countries-config --from-file=countriesConfig.yaml -o yaml --dry-run \
```
galund: Maybe this should be a separate script as then it could be commented (I have no idea what the `num-versions` stuff is about.

ci/build/build-pipeline.yaml
Wynndow: Nifty! I'll update

ci/sandbox/build-pipeline.yaml
Wynndow: As in, a separate script in a file somewhere?

Wynndow: Also, it's this: https://github.com/k14s/kapp/blob/master/docs/diff.md#versioned-resources

galund: Yeah, in a file was what I was thinking about. I understand now, sorta, what the version resources thing is, but I don't know how we'd end up deciding that 5 is the right magic number to keep :-)

Wynndow: To be honest I just picked a sensible sounding number, and then later realised that 5 is actually the default. I left this is to be explicit. I think I'm gonna leave this as it is with regard to moving it to a new file/commenting. There's a link to the docs in the commit message if people are interested in what it's doing.

galund (dismissed):

galund (approved):

JohnWattsGDS (approved):

PR #410, opened 2019-12-03
🚬EID-1854 Smoke Test Infrastructure

idrop:
https://govukverify.atlassian.net/browse/EID-1854
Introduce infrastructure to support smoke testing our integration and production releases.

The idea of this PR is to reuse the existing harbor image we build for acceptance tests to also run smoke tests.

Therefore in our build pipeline we target only the acceptance test cucumber feature, plus introduce a new smoke test cucumber feature, which will be run in the new smoke-test pipeline.
The pipelines now invoke a cucumber test either by:
* `bundle exec cucumber features/proxy_node.feature` or
* `bundle exec cucumber features/smoke_test.feature`

This idea is being tested out on the release branch, as it has an integration release.

The integration smoke test job is triggered every 10 minutes. There are currently no consequences on failure.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -621,7 +621,7 @@ spec:
             - -euc
             - |
               cd /
-              bundle exec cucumber --strict --tags "not @ignore"
+              bundle exec cucumber features/proxy_node.feature --strict --tags "not @ignore"
 
     - name: release
       serial: true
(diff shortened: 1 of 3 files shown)


Wynndow (approved):
I guess in future we could have different features per environment - so in integration we could do a full journey with stub-idp but in prod we could just touch the hub.

Or would that be more than just a "smoketest" in integration?

JohnWattsGDS (approved):

PR #416, opened 2019-12-12
BAU: Local running against a locally running Verify Hub

CharlesIC:
Add configuration to local startup to allow sending requests to a locally running Hub.

To allow talking to a local Hub, we need an alternative VSP config - since that's the component that generates AuthRequests to the Hub.

The changes needed are:
  - Custom config file with no `environment` set and more `verifyHubConfiguration` overrides
  - More environment variables in the `docker.env` file to include the new overrides
  - New Dockerfile to use the custom config
  - New private signing and encryption keys and certs - these certs are already trusted by the Hub for local startup (reused test RP keys and certs)
  - New trustore with the local Hub metadata signing cert so the VSP can trust the local Hub
  - New service in docker-compose to run the VSP locally with the alternative config: `verify-service-provider-local-hub`
  - New flag in the local startup script to specify the local Hub config

Note that only one VSP should be run at any one time - either with the regular compliance tool config or with now the new local hub config.

The Proxy Node config for the local Hub was added in alphagov/ida-hub-acceptance-tests#161


idrop (approved):
works locally for me 👍

galund (approved):

PR #417, opened 2019-12-17
BAU: Add a killswitch to integration pipeline

idrop:
Helps with fixing HSM key creation issues.

Provides a way to delete deployments, virtual services for integration releases, making it easier to diagnose remaining interactions with the HSM.

Does not require RE to assume admin role to delete things. We can redeploy a release by starting a job on the integration deploy pipeline (all currently paused).

This new pipeline is a copy of the prod killswitch, with a job name change.

There may be another PR to killswitch the VMC.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,7 +5,7 @@ ignore:
   SNYK-JAVA-CAJULIUSDAVIES-30073:
     - '*':
         reason: Fix not available
-        expires: 2019-12-15T00:00:00.000Z
+        expires: 2020-01-15T00:00:00.000Z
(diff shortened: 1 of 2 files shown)


CharlesIC (dismissed):

galund (approved):
Might be good to de-duplicate somehow but 👍

Wynndow:
Will this remove the VMC deployment? If the intention of this is to remove all connections with the HSM it'll need to also wipe out the VMC. It may well already do this, my brain is struggling to work it out though.

CharlesIC (approved):

PR #419, opened 2019-12-17
EID-1872 Delete test release nightly & before test

idrop:
The test connector in particular is creating handles to HSM keys which is contributing to us going over quota.

We want to reap these key handles by regularly deleting the test release.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -176,15 +176,23 @@ spec:
         <<: *harbor_source
         repository: registry.((cluster.domain))/eidas/verify-service-provider
 
+    - name: nightly
+      type: time
+      icon: update
+      source:
+        interval: 1h
+        start: 3:00 AM
+        stop: 4:00 AM
+
     groups:
       - name: deployment
-        jobs: [build-vsp-image, build-proxy-node, package, test, release]
+        jobs: [build-vsp-image, build-proxy-node, package, test, release, delete]
       - name: proxy-node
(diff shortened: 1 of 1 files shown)


ci/build/build-pipeline.yaml
```diff
@@ -510,7 +558,7 @@ spec:
       - in_parallel:
           steps:
           - get: chart
-            passed: [package]
+            passed: [delete]
```
CharlesIC: This is weird.... We're testing a package whose deployment has been deleted?
Or is this intentional? We're always deleting the test instance before testing?

CharlesIC: If the latter is the case - just add this task to the test job - no need for a new job here

Wynndow: Yep, always deleting the previous test instance before recreating and testing. Ensures we're starting from a new release for all the components.

CharlesIC: Ok then the `task: delete-test-release` logically belongs to the test job - clean up before releasing a new test instance

Wynndow: The plan was to have a separate job so we can re-use the logic. This will be triggered nightly, as well as before any test release. That way the test environment will be left after a test run (which may be useful for debugging), but will never live for more than 24 hours, to deal with the HSM key issue.

CharlesIC: All right then, I guess that makes sense

PR #420, opened 2019-12-17
EID-1873 Annotate deployments to ensure nightly release

Wynndow:
I'm not 100% sure this will work, looking for opinions.

We release the apps nightly on a trigger in Concourse. However kapp's
behaviour is to only release stuff if it's changed. If nothing has
changed, kapp won't actually re-deploy apps for us. You can see an
example of this [here](https://ci.london.verify.govsvc.uk/teams/proxy-node-prod/pipelines/deploy/jobs/deploy-cz-production/builds/122).This may be
contributing to the issue with the HSM retaining keys.

We should be able to annotate these deployments with a timestamp. This
will mean that kapp detects a change in the app when in renders the
manifest and will redeploy it.


Wynndow:
Closing in favour of https://github.com/alphagov/verify-proxy-node/pull/421

PR #432, opened 2019-12-20
BAU - Bump cloudhsm tools to 3.0.0

Wynndow:
There is a new major version of the [cloudhsm client software](https://docs.aws.amazon.com/cloudhsm/latest/userguide/client-history.html)

I was looking into this to see if they had released a fix to the issue
we're experiencing with the hsm retaining temporary keys. It looks like
this version does not fix that issue, but it's still probably a good
idea to upgrade.

I've applied this change to the sandbox without any issues.


Wynndow:
yep, I did notice them in the VMC - will do that next

PR #435, opened 2020-01-02
Replace concourse.k8s.io API group with concourse.govsvc.uk

Krenair:
See https://github.com/alphagov/gsp/pull/858


samcrang:
The build on this failing because of Snyk and some vulnerable Java dependencies

PR #436, opened 2020-01-03
EID-1877 Decouple Stub Connector from the HSM

idrop:
_TLDR; The Stub Connector represents a stub country. It should not have its saml signing coupled to an HSM key._

---

See https://govukverify.atlassian.net/browse/EID-1877

The Stub Connector (a stub country) currently has its metadata certs+keys generated via the verify-metadata-controller, and the saml signing key is held in the HSM, see https://test-integration-connector.london.verify.govsvc.uk/ConnectorMetadata

Every SAML request from stub connector is currently signed by a key in the CloudHSM. This was done as a convenience as the truststore for this signing is also generated via verify-metadata-controller, and easily made available to the ESP (the service that reads the connector metadata).

We want to model a a stub country that has its own PKI completely outside the HSM. The eventual goal is to run the Stub Connector in PaaS, to complete the separation.

Also, the Stub Connector is holding on to HSM key handles when it generates SAML requests.
This contributes to our quota of a max of ~ 3300 handles.

This PR will:

* Create a pipeline to set a metadata signing cert+key, saml signing cert+key plus a metadata signing truststore, all from verify-dev-pki, into a cluster-wide configmap named test-pki-configmap
* Read test-pki-configmap to set a truststore in ESP
* Read test-pki-configmap via a volumeMount to provide metadata and saml signing credentials to the connector
* Remove HSM client config from stub connector
* Publish connector node metadata from a Resource using a mustache template. The method is equivalent to how connector node metadata is currently created in the verify-metadata-controller
* Use sensible default values in connector config.yml to avoid setting config via env vars

You can see connector node metadata created with a cherry pick of this commit at https://stub-connector.eidas.test.london.sandbox.govsvc.uk/ConnectorMetadata

Out of Scope: 
* Create two instances of stub country, on the master branch
* Remove current process for publishing metadata within verify-metadata-controller
* Publish the metadata signing CA certs, see https://test-integration-connector.london.verify.govsvc.uk/ConnectorMetadataSigningCertificates
* Migrate to PaaS

### DIFF
--- a/chart/templates/esp-deployment.yaml
+++ b/chart/templates/esp-deployment.yaml
@@ -70,14 +70,11 @@ spec:
 {{- if .Values.stubConnector.enabled }}
         - name: CONNECTOR_NODE_METADATA_TRUSTSTORE
           valueFrom:
-            secretKeyRef:
-              name: {{ .Release.Name }}-connector-metadata
-              key: metadataSigningTruststoreBase64
+            configMapKeyRef:
+              name: test-pki-configmap
+              key: stub_connector.truststore.base64
         - name: CONNECTOR_NODE_METADATA_TRUSTSTORE_PASSWORD
-          valueFrom:
-            secretKeyRef:
-              name: {{ .Release.Name }}-connector-metadata
-              key: metadataSigningTruststorePassword
+          value: {{ .Values.connector.metadataSigningTruststorePassword }}
 {{- else }}
         - name: CONNECTOR_NODE_METADATA_TRUSTSTORE
           value: {{ .Values.connector.metadataSigningTruststoreBase64 }}
--- a/ci/build/test-pki.yaml
+++ b/ci/build/test-pki.yaml
@@ -27,7 +27,6 @@ spec:
     - name: build-test-pki
       plan:
       - get: verify-dev-pki
-        trigger: true
       - task: assemble-pki
         config:
           platform: linux
--- a/ci/build/test-pki.yaml
+++ b/ci/build/test-pki.yaml
@@ -1,0 +1,83 @@
+---
+apiVersion: concourse.k8s.io/v1beta1
+kind: Pipeline
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: build-test-pki
+spec:
+  exposed: true
+  config:
+
+    task_toolbox: &task_toolbox
+      type: docker-image
+      source:
+        repository: ((concourse.task-toolbox-image))
+        tag: ((concourse.task-toolbox-tag))
+
+    resources:
+
+    - name: verify-dev-pki
+      type: git
+      source:
+        uri: https://github.com/alphagov/verify-dev-pki.git
+
+    jobs:
+
+    - name: build-test-pki
+      plan:
+      - get: verify-dev-pki
+        trigger: true
+      - task: assemble-pki
+        config:
+          platform: linux
+          image_resource:
+            type: docker-image
+            source:
+              repository: openjdk
+              tag: 11
+          inputs:
+            - name: verify-dev-pki
+          outputs:
+            - name: pki
+          run:
+            path: /bin/sh
+            args:
+              - -euc
+              - |
+                echo "create a truststore containing metadata-ca and ida-root-ca"
+                keytool -keystore ./stub_connector.truststore -import -v -file verify-dev-pki/src/main/resources/ca-certificates/ida-root-ca.pem.test -storepass marshmallow -trustcacerts -noprompt -alias ida-root-ca.pem.test
+                keytool -keystore ./stub_connector.truststore -import -v -file verify-dev-pki/src/main/resources/ca-certificates/ida-metadata-ca.pem.test -storepass marshmallow -trustcacerts -noprompt -alias ida-metadata-ca.pem.test
+                openssl base64 -A -in stub_connector.truststore > pki/stub_connector.truststore.base64
+
+                echo "copy keys and certs to pki/"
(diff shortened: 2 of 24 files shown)


stub-connector/src/main/java/uk/gov/ida/notification/stubconnector/resources/ConnectorNodeCredentialConfiguration.java
```diff
@@ -0,0 +1,71 @@
+package uk.gov.ida.notification.stubconnector.resources;
```
Wynndow: Is this the correct place for this? Perhaps somewhere like `uk.gov.ida.notification.stubconnector.configuration`

stub-connector/src/main/java/uk/gov/ida/notification/stubconnector/resources/MetadataResource.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private final MetadataGenerator metadataGenerator;
+
+    public MetadataResource(MetadataGenerator metadataGenerator) {
+        this.metadataGenerator = metadataGenerator;
+    }
+
+    @GET
+    public Response getConnectorNodeMetadata() throws Exception {
+        XMLObject xmlObject = metadataGenerator.getConnectorMetadata();
```
Wynndow: Rather than regenerate and sign the metadata every time this endpoint is hit, could it be cached? Or just generate it once when the `MetadataResource` is initialised, with a really long validity period?

chart/templates/esp-deployment.yaml
```diff
@@ -70,14 +70,11 @@ spec:
 {{- if .Values.stubConnector.enabled }}
         - name: CONNECTOR_NODE_METADATA_TRUSTSTORE
           valueFrom:
-            secretKeyRef:
-              name: {{ .Release.Name }}-connector-metadata
-              key: metadataSigningTruststoreBase64
+            configMapKeyRef:
```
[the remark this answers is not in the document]
idrop: Mainly so that we can see exactly how the truststore, plus keys and certs were created.
The fact that the configmap resolves to files in the `/pki` directory when volume mounted makes it even more attractive.
Granted this would be obsolete if/when we move to PaaS.

stub-connector/src/main/java/uk/gov/ida/notification/stubconnector/resources/ConnectorNodeCredentialConfiguration.java
idrop: Agree! Moving

stub-connector/src/main/java/uk/gov/ida/notification/stubconnector/resources/MetadataResource.java
idrop: As this is the stub connector I wasn't too concerned with caching, but I'm going to add a `@CacheControl` annotation to simulate a cache. Also adding a metadata expiry env var so we can simulate a stub country with expired metadata.

Wynndow: I guess I was thinking more that the metadata isn't going to change or be regenerated once the app is up and running, so to save processing does it make sense to just generate the metadata once when the app boots up, rather than doing it every time the endpoint is hit. The cache you've added will help but it might be better to just only do it once but with a long validity.

Having said this, the amount it'll get used means this probably doesn't matter at all.

proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/ResponseAssertionDecrypter.java
```diff
@@ -18,10 +16,6 @@
 
     public ResponseAssertionDecrypter(Credential credential) {
         this.decrypter = new DecrypterFactory().createDecrypter(Collections.singletonList(credential));
-        if (credential.getEntityId() != null && credential.getEntityId().equals(CloudHsmCredentialConfiguration.ID)) {
-            ProxyNodeLogger.info("Using CloudHSM so set JCA provider to Cavium");
-            this.decrypter.setJCAProviderName("Cavium");
```
galund: Are we sure we no longer need this? The translator also uses the `ResponseAssertionDecrypter` (maybe)?

ci/build/test-pki.yaml
```diff
@@ -0,0 +1,83 @@
...
+      type: git
+      source:
+        uri: https://github.com/alphagov/verify-dev-pki.git
+
+    jobs:
+
+    - name: build-test-pki
+      plan:
+      - get: verify-dev-pki
+        trigger: true
```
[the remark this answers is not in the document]
idrop: Removed the trigger, we can invoke by hand

ci/sandbox/test-pki.yaml
```diff
@@ -0,0 +1,85 @@
...
+      type: git
+      source:
+        uri: https://github.com/alphagov/verify-dev-pki.git
+
+    jobs:
+
+    - name: build-test-pki
+      plan:
+      - get: verify-dev-pki
+        trigger: true
```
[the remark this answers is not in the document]
idrop: Removed the trigger, we can invoke by hand

proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/ResponseAssertionDecrypter.java
galund: (Nah, that's only in a test, and has a non-HSM key as well.)

galund (approved):

Wynndow (approved):

PR #443, opened 2020-01-13
BAU Slack alert on a pipeline job failure

idrop:
Send a slack message to `#verify-eidas-alerts` when a concourse job fails on the
* [build](https://ci.london.verify.govsvc.uk/teams/proxy-node-build/pipelines/build-release),
* [integration deploy](https://ci.london.verify.govsvc.uk/teams/proxy-node-integration/pipelines/deploy),
* [integration smoke](https://ci.london.verify.govsvc.uk/teams/proxy-node-integration/pipelines/smoke-test) or 
* [prod deploy](https://ci.london.verify.govsvc.uk/teams/proxy-node-prod/pipelines/deploy) pipelines.

The slack webhook url required by the resource is referenced in each pipeline via the `((slack-notifications.url))` var. The value of this url is set via a sealed secret per namespace.

Secrets are sealed [using this method](https://verify-team-manual.cloudapps.digital/documentation/support/eidas-proxy-node/connecting-vsp-to-hub.html#creating-a-sealed-secret-for-the-vsp).

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -54,6 +54,14 @@ spec:
         repository: amazoncorretto
         tag: 11
 
+    slack-on-failure: &slack-on-failure
+      on_failure:
+        put: notify
+        params:
+          text: |
+            Proxy Node pipeline failed on the `$BUILD_PIPELINE_NAME.$BUILD_JOB_NAME` step.
+            https://ci.london.verify.govsvc.uk/teams/proxy-node-build/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME
+
     resource_types:
 
     - name: github
@@ -68,6 +76,11 @@ spec:
       source:
         repository: ((concourse.harbor-resource-image))
         tag: ((concourse.harbor-resource-tag))
+
+    - name: slack-notification
+      type: docker-image
+      source:
+        repository: cfcommunity/slack-notification-resource
 
     resources:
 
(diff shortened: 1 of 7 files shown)


ci/build/build-pipeline.yaml
```diff
@@ -54,6 +54,14 @@ spec:
...
         tag: 11
 
+    slack-notification: &slack-notification
+      on_failure:
+        put: notify
+        params:
+          text: |
+            Proxy Node pipeline failed on the `$BUILD_PIPELINE_NAME.$BUILD_JOB_NAME` step.
+            https://ci.london.verify.govsvc.uk/teams/proxy-node-build/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME
+
```
Todai88: nit: I personally prefer to create the anchor where I first use it.

ci/build/build-pipeline.yaml
```diff
@@ -199,6 +217,7 @@ spec:
     - name: build-vsp-image
       serial: true
       serial_groups: [build-vsp]
+      <<: *slack-notification
```
Todai88: The more I read about the `<<:` operator, the less convinced I am that it's necessary. 

`<<:` seems to perform a merge, so it's useful when you want to add some key-value-pairs from an anchor, but without overwriting existing keys. We don't have any existing keys, so it isn't to perform a merge necessary. 

a `*` should be enough.

That being said, what you have will work as intended.

Todai88 (dismissed):
Looks good. I've added some nits that you can address, if you want to.

ci/build/build-pipeline.yaml
idrop: My thinking was if the anchor is defined at the top of the yaml, its not ambiguous in the pipeline which of `&` or `*` is the anchor or alias 🤷‍♂

ci/build/build-pipeline.yaml
idrop: Agree, I've pushed a change to remove the << chevrons

PR #445, opened 2020-01-16
EID-1873 add timestamp annotation on pod

idrop:
Add a timestamp annotation on the spec template during helm templating.
This will mark the pod referenced in the deployment as changed, and k8s should reconcile with a redeploy.
This is required for nightly rolling.


chart/templates/stub-connector-deployment.yaml
```diff
@@ -25,6 +23,8 @@ spec:
         app.kubernetes.io/name: connector
         app.kubernetes.io/instance: {{ .Release.Name }}
         talksToHsm: "true"
```
idrop: I'm going to remove this is a separate PR, the stub-connector no longer talks to the HSM

PR #446, opened 2020-01-17
EID-1887: Create DTO for Metatron response

Wynndow:
https://govukverify.atlassian.net/browse/EID-1887

Adds a DTO for use by Metatron and the client the apps will be using to fetch details from metadata for different countries.


JohnWattsGDS (dismissed):

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/MetatronResponse.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.net.URI;
+
+public class MetatronResponse {
+
+    private String samlSigningCertX509;
+    private String samlEncryptionCertX509;
+    private URI destination;
+    private String entityId;
+    private String countryCode;
```
idrop: Should we annotate these fields with validators? We're doing it for other DTOs in this package. In this way we'd get a useful 400 error response in clients.

Wynndow: I was wondering the same thing, but have had doubts for a while that we actually use these annotations properly elsewhere. If you look at the first commit of this PR I wrote some tests to check their behaviour. These validators only automatically come into play if you are using the DTO as the parameter type for a resource. If you're using it to deserialize a response (as we will be) they don't. _Unless_ you specifically new up and invoke a validator. It seems like the reasoning is to make it easy to validate stuff you're receiving from the outside world.

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/MetatronResponse.java
```diff
@@ -0,0 +1,38 @@
+package uk.gov.ida.notification.contracts;
+
+import java.net.URI;
+
+public class MetatronResponse {
```
CharlesIC: I know this is a cool name but maybe something like `CountryMetadataInfo` would be more descriptive of what this DTO represents?

Wynndow: Perhaps, but it is in keeping with the naming conventions we've used for other DTO's in the proxy node, like `EidasSamlParserResponse` and `HubResponseTranslatorRequest`

CharlesIC: I don't think those are particularly good names so would call this new DTO whatever is the best name, which I think is something specific and more descriptive

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/MetatronResponse.java
Wynndow: Phil and I just dug into this. The validators can be used on the server side by annotating the resource with `@Valid`. If the object being serialized isn't valid a 500 is returned. I'm going to add validations and the `@Valid` annotation to this DTO.

CharlesIC (dismissed):

JohnWattsGDS (dismissed):
I'm pretty sure MetatronResponse is a superior name for this class but approving anyway ;)

CharlesIC (approved):

idrop (approved):

PR #447, opened 2020-01-20
EID-1833: Create a shared client for Metatron

Wynndow:
This PR has been done before the endpoint on the Metatron the client will be accessing [has been created](https://govukverify.atlassian.net/browse/EID-1828), so it's made a couple of assumptions. First, that the new endpoints URL will be taking the countries entity ID as a path parameter. Second, I've URL encoded the entity ID to allow it to be used in the URL. If this isn't the approach taken this PR will need adjusting.


Wynndow:
Don't review yet, need to update with the new DTO's name.

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/proxy/MetatronResponseException.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public MetatronResponseException(Throwable cause, String entityId) {
+        super(String.format(ERROR_MESSAGE_FORMAT, entityId), cause);
+        MDC.put(ProxyNodeMDCKey.EIDAS_ENTITY_ID.name(), entityId);
+    }
+
+    @Override
+    public Response.Status getResponseStatus() {
+        ApplicationException cause = (ApplicationException) this.getCause();
+        return cause.getExceptionType() == ExceptionType.CLIENT_ERROR ?
```
idrop:
```suggestion
        return ExceptionType.CLIENT_ERROR == cause.getExceptionType() ?
```

idrop: To avoid a potential NPE, happy to be be proven that it's not possible

build.gradle
```diff
@@ -34,7 +34,7 @@ repositories {
 ext {
     opensaml_version = '3.4.3'
     dropwizard_version = '1.3.16'
-    utils_version = '2.0.0-371'
+    utils_version = '2.0.0-375'
```
idrop: How do you know to bump to latest versions? I wouldn't have thought to do this.

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/MetatronProxy.java
```diff
@@ -0,0 +1,39 @@
...
+    private final URI metatronUri;
+
+    public MetatronProxy(
+            ProxyNodeJsonClient metatronClient,
+            URI metatronUri) {
+        this.metatronClient = metatronClient;
+        this.metatronUri = metatronUri;
+    }
+
+    public CountryMetadataResponse getCountryMetadata(String entityId) {
```
idrop: It bugs me that entityId is a String when it could possibly be a URI, but its used as a String is throughout our codebase. Its a me problem 😢

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/MetatronProxyTest.java
```diff
@@ -0,0 +1,144 @@
...
+        return new ProxyNodeJsonClient(
+                new ErrorHandlingClient(ClientBuilder.newClient()),
+                new JsonResponseProcessor(new ObjectMapper()),
+                new IstioHeaderStorage()
+        );
+    }
+
+    @Path("/metadata")
+    @Produces(MediaType.APPLICATION_JSON)
+    public static class TestMetatronResourceGoodResponse {
```
idrop: I ❤️ that these test resources are in the test class 😄

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/proxy/MetatronResponseException.java
Wynndow: Both sides of the expression are going to be evaluated either way I think. Not sure it would cause a NPE though? An `ApplicationException` can't be instantiated without an exception type being provided, and even if it was provided a null, `null == ExceptionType.CLIENT_ERROR` is still a valid comparison.

build.gradle
Wynndow: Ah, I had to update the rest-utils library in this PR to allow a GET request with added headers. This was to allow the client to make a request and pass Istio headers along like we do elsewhere. https://github.com/alphagov/verify-utils-libs/pull/49

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/MetatronProxy.java
Wynndow: Yeah, I keep thinking the same thing! I think it boils down to that we don't actually use the entityId as URL, just as a label or identifier, so we don't need all the bells and whistles that come with having a URI.

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/MetatronProxyTest.java
Wynndow: There is probably a fancier way of doing it. This seems to work just fine though!

proxy-node-shared/src/main/java/uk/gov/ida/notification/exceptions/proxy/MetatronResponseException.java
idrop: You're right, let's keep it as is

proxy-node-shared/src/test/java/uk/gov/ida/notification/shared/proxy/MetatronProxyTest.java
JohnWattsGDS: At some iteration, I'll copy this and put the metatron resources and associated gubbins inside the test classes.  It seems a much smarter / easier to understand way of doing it.

PR #450, opened 2020-01-22
EID-1827 fetch countries metadata

JohnWattsGDS:
Use verify-saml-libs 232 and apply the changes in https://github.com/alphagov/verify-proxy-node/pull/412
    
Tidy up unused imports.
    
Replace some sysouts with ProxyNodeLogger
    
MetadataResolverService to retrieve and cache metadata from the URLs provided in EidasConfig.

Add "metadata" endpoint as temporary endpoint for use in tests.  End point will currently return a reponse with status 200 and the metadata entity id for successfully resolved metadata and 500 if there are any problems.  This will be replaced with the endpoint returning the MetatronResponse [sic] DTO and the tests will be updated accordingly in future.

Provide initial test classes.  The CreatesConfigDropwizardClientRule writes the countriesConfig.yaml used by MetatronAppRule to /tmp and makes the temporary port that it will listen on available.

TestMetadataResource uses the getPort method of CreatesConfigDropwizardClientRule in setting the entityId of the metadata resources it creates.

MetatronTests tests that the responses are as expected for the various metadatas gathered by Metatron(AppRule).  When the correct test-metadatas are provided to TestMetadataResource then metatron will successfully resolve two of them and fail to resolve / provide metadata that is, unsigned, expired or untrusted.


Wynndow:
There's a lot of commits here and a big merge conflict. Think it might need a rebase?

JohnWattsGDS:
Yeah, I just noticed.  I guess that's what I get for trying to squash the commits that were there ...

libs/metatron/pom.xml
```diff
@@ -54,6 +54,90 @@
         </exclusion>
       </exclusions>
     </dependency>
```
Wynndow: This `libs` dir and contents can be removed. The gradle task that creates these poms was [recently removed](https://github.com/alphagov/verify-proxy-node/commit/d011c607376ec493329e312e520cb02fdbea9d7f) - suspect this was generated and commited before that happened.

metatron/build.gradle
```diff
@@ -4,8 +4,21 @@ version '1.0-SPIKE'
 dependencies {
     compile configurations.dropwizard,
             configurations.common,
+            configurations.verify_saml,
+            project(':proxy-node-shared'),
+            'com.sun.xml.bind:jaxb-core:2.3.0',
+            'com.sun.xml.bind:jaxb-impl:2.3.0',
+            "org.opensaml:opensaml-core:$opensaml_version",
```
Wynndow: A lot of these opensaml deps are already specified in a configuration in the `build.gradle` in the root and could be replaced here with `configurations.opensaml`. Suspect some of the others may be defined as well.

libs/metatron/pom.xml
JohnWattsGDS: 👍

metatron/build.gradle
JohnWattsGDS: 👍

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetatronConfiguration.java
```diff
@@ -1,7 +1,18 @@
...
 import io.dropwizard.Configuration;
 
-public class MetatronConfiguration extends Configuration {
+public class MetatronConfiguration extends Configuration implements PrometheusConfiguration {
 
+    private String countriesConfig;
+
+    @Override
+    public boolean isPrometheusEnabled() {
+        return true;
```
Wynndow: Should this be over-rideable, for running locally and what-not?

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetricsUtils.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.ida.eidas.metatron;
+
+public class MetricsUtils {
```
Wynndow: I can't find this class being used anywhere - is it needed for some sort of prometheus magic?

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasConfig.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+
+public class EidasConfig {
+    @JsonProperty
```
Wynndow: What's this annotation for? I've seen it being used like this but wasn't sure what using it without an argument did?

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetatronConfiguration.java
JohnWattsGDS: Well, strictly speaking since I took the Prometheus thing out (as it was in verify-saml-libs anyway), it's not needed at all but it's a straight copy from somewhere else and if Prometheus is enabled, it just adds a servlet to the container so wouldn't I think affect any thing running locally.

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetricsUtils.java
JohnWattsGDS: Again, this is now redundant.  But well done on spotting things Codacy doesn't.  Take that as you will ;)

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public EidasCountryConfig getCountry(String country) {
+        if(Objects.isNull(countriesMap)) {
+            countriesMap = new HashMap<>();
+            countries.stream().forEach(eidasCountryConfig -> countriesMap.put(eidasCountryConfig.getName(), eidasCountryConfig));
+        }
+        return countriesMap.get(country);
+    }
+
+    public void retainAll(Function<EidasConfig, Collection<EidasCountryConfig>> filterFunction) {
```
Wynndow: I can't see this method being used anywhere. Is it needed? What does it do?

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasConfig.java
JohnWattsGDS: I see it as a marker annotation without any arguments; as in making it clear that this file comes from some json at some point and that this is the 'thing' that gets populated by the json.  So it's an annotation in the simplest sense of the word.

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/KeyStoreModule.java
```diff
@@ -48,11 +49,10 @@ public KeyStore deserialize(JsonParser jp, DeserializationContext ctxt) throws I
             try {
                 byte[] decoded = Base64.getDecoder().decode(node.asText());
                 keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
-                keyStore.load(new ByteArrayInputStream(decoded), "marshmallow".toCharArray());
+                keyStore.load(new ByteArrayInputStream(decoded), "".toCharArray());
```
Wynndow: Does the config tool create the truststores with an empty password?

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasConfig.java
JohnWattsGDS: This is a hangover from the config repo where this class was born, this and the other method can go as they're not used here.

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/KeyStoreModule.java
```diff
@@ -48,11 +49,10 @@ public KeyStore deserialize(JsonParser jp, DeserializationContext ctxt) throws I
...
             } catch (KeyStoreException | IOException | NoSuchAlgorithmException | CertificateException e) {
                 keyStore = null;
```
Wynndow: Could this be moved to where the keystore is declared on line 48 (`KeyStore keyStore = null`) to save having this line here?

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasConfig.java
```diff
@@ -0,0 +1,36 @@
...
+    private List<EidasCountryConfig> countries = new ArrayList<>();
+
+    @JsonIgnore
+    private Map<String, EidasCountryConfig> countriesMap;
+
+    public Collection<EidasCountryConfig> getCountries() {
+        return countries;
+    }
+
+    public EidasCountryConfig getCountry(String country) {
```
Wynndow: This doesn't seem to be used either - the resolver service just pulls the countries from the `countries` list - can it be dropped?

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -0,0 +1,78 @@
...
+        this.metadataResolverMap.keySet().stream().forEach(k -> metadataHealth.add(new MetadataHealth(k, (AbstractReloadingMetadataResolver) metadataResolverMap.get(k))));
+        return metadataHealth;
+    }
+
+    public MetadataResolver getMetadataResolver(String entityId) {
+        return metadataResolverMap.get(entityId);
+    }
+
+    private Client getClient(EidasCountryConfig country) {
+        if (Objects.isNull(country.getTlsTruststore())) {
```
Wynndow: Maybe `getTlsTruststore()` should be returning an optional?

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/KeyStoreModule.java
JohnWattsGDS: I guess it could, but even if we were paying for storage by the byte I feel like it's worth extra few characters to point out that we're deliberately returning a null keyStore (which I'm not in love with) rather than that we're ignoring the fact that it failed to create one.

metatron/src/main/java/uk/gov/ida/eidas/metatron/health/MetadataHealth.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private String entityId;
+    private DateTime expirationTime;
+    private DateTime lastRefresh;
+    private DateTime lastSuccessfulRefresh;
+    private boolean wasLastRefreshSuccessful;
+    private DateTime nextRefresh;
+
+    public MetadataHealth(String entityId, AbstractReloadingMetadataResolver resolver) {
+        this.entityId = entityId;
```
Wynndow: Rather than having all the elements as separate fields, could you just have `entityId` and `resolver`, and then have the getters call methods on the resolver? I know that at the moment the methods are being called immediately after the object is created, so it doesn't matter much, but if this was ever used somewhere else where the object is created a while before it's accessed, you might end up with stale data.

metatron/src/main/java/uk/gov/ida/eidas/metatron/health/MetatronHealthCheck.java
```diff
@@ -3,9 +3,8 @@
 import com.codahale.metrics.health.HealthCheck;
 
 public class MetatronHealthCheck extends HealthCheck {
-
     @Override
     protected Result check() {
-        return Result.healthy("I'm completely operational, and all my circuits are functioning perfectly");
+        return Result.healthy();
```
Wynndow: I guess this could be extended later to query the resolvers to see if enough of them are happy.

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
JohnWattsGDS: I think that makes the relatively simple json to pojo conversion a little more complex at the EidasCountryConfig level and we'd still end up having to do some kind of check on whether the Optional was empty or not.  As this method is only used once (per client created) when the ResolverService is spun up I think it's okay to have it as it is.  If it were used as part of a public API I think there's more of a case for it but in this situation it's okay.

metatron/src/main/java/uk/gov/ida/eidas/metatron/resources/MetatronResource.java
```diff
@@ -1,35 +1,50 @@
...
                 .map(EidasCountryConfig::toString)
-                .reduce("", (in, config) -> in + config) ;
+                .reduce("", (in, config) -> in + config);
     }
 
     @GET
-    @Path("/truststore")
-    public String truststore() throws KeyStoreException {
-        KeyStore truststore = config.getKeyStore();
+    @Path("/metadata-truststore/{country}")
```
Wynndow: This route is for testing right? We'll probably remove it before Metatron is released to the wild?

metatron/src/main/java/uk/gov/ida/eidas/metatron/resources/MetatronResource.java
```diff
@@ -43,6 +58,28 @@ public String truststore() throws KeyStoreException {
         });
         return trustedCerts.toString();
     }
+
+    @GET
+    @Path("/metadata/{entityId}")
+    public String metadata(@PathParam("entityId") String entityId) throws ResolverException {
```
Wynndow: This and the `last-refresh` route are for testing too?

JohnWattsGDS: Yeah, this and the are just a temporary things so we have something to poke whilst we're bringing it into something useful.  I guess if we use Prometheus, there might only be one endpoint that returns the MetatronResponse [sic].  At the moment, the entityId is the easiest thing to resolve without lots of extra code.

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/CreatesConfigDropwizardClientRule.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.Assert;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.net.URI;
+
+public class CreatesConfigDropwizardClientRule extends DropwizardClientRule {
+    public final static String CONFIG_LOCATION = "/tmp/countriesConfig.yaml";
```
Wynndow: Might be worth using [createTempDirectory()](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory(java.lang.String,%20java.nio.file.attribute.FileAttribute...)) - potentially more portable as it apparently uses the systems default temp dir.

metatron/src/main/java/uk/gov/ida/eidas/metatron/health/MetadataHealth.java
JohnWattsGDS: I don't expect this to live too long.  It's useful for poking and meeting the original acceptance criteria but I'm pretty sure it's going to be superseded by Prometheus which will work completely differently.

Having a reference to a resolver would make more sense but I suspect that once we start retrieving actual metadata data there will be an extra layer of abstraction that will pretty much do what you're saying.

metatron/src/main/java/uk/gov/ida/eidas/metatron/health/MetatronHealthCheck.java
JohnWattsGDS: There's quite a bit of cool stuff we could do here.

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/CreatesConfigDropwizardClientRule.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public CreatesConfigDropwizardClientRule(MetatronTests.TestMetadataResource testMetadataResource) {
+        super(testMetadataResource);
+        try {
+            this.before();
```
Wynndow: Did you find out what `this.before` was doing?

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/CreatesConfigDropwizardClientRule.java
JohnWattsGDS: Good suggestion.

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/CreatesConfigDropwizardClientRule.java
JohnWattsGDS: The port was coming back as 0 unless I used that.

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/CreatesConfigDropwizardClientRule.java
```diff
@@ -0,0 +1,41 @@
...
+        } catch (IOException e) {
+            Assert.fail(e.getMessage());
+        }
+    }
+    private void writeConfigFile() throws IOException {
+        URI uri = super.baseUri();
+        String templateFilePath = this.getClass().getClassLoader().getResource("countriesConfig.yaml").getFile();
+        String templateFile = new String(new FileInputStream(new File(templateFilePath)).readAllBytes());
+        String fileWithHost = templateFile.replace("{host_holder}", uri.toString());
+        FileWriter writer = new FileWriter("/tmp/countriesConfig.yaml");
```
Wynndow: Can use `CONFIG_LOCATION` here

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/CreatesConfigDropwizardClientRule.java
```diff
@@ -0,0 +1,41 @@
...
+        } catch (Throwable throwable) {
+            Assert.fail(throwable.getMessage());
+        }
+        try {
+            writeConfigFile();
+        } catch (IOException e) {
+            Assert.fail(e.getMessage());
+        }
+    }
+    private void writeConfigFile() throws IOException {
```
Wynndow: Nice...

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
```diff
@@ -0,0 +1,263 @@
...
+        waitWhile(() -> testMetadataServer.getPort() == 0, "Waiting for client rule port ");
+
+        testMetadataResource.init(testMetadataServer.getPort());
+        metatronAppRule = new MetatronAppRule();
+
+        return RuleChain.outerRule(testMetadataServer).around(metatronAppRule);
+    }
+
+    private static void waitWhile(Supplier<Boolean> condition, String message) {
+        // This needs a time-out mechanism
```
Wynndow: Could create a datetime, and then compare the current datetime to it in the loop. If it's more than 1 second or two, bail out?

Wynndow: Also, system out printing doesn't appear anywhere that I can see, could drop it?

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
```diff
@@ -0,0 +1,263 @@
...
+    private String getEntityId(String name) {
+        return "http://localhost:" + testMetadataServer.getPort() + "/application/" + name + "/Metadata";
+    }
+
+    private String getUriString(String entityId) {
+        return "http://" + UriBuilder.fromMethod(MetatronResource.class, "metadata").host("localhost").port(metatronAppRule.getLocalPort()).build(entityId).toString();
+    }
+
+    private Client getClient() {
+        if ( client == null ) {
```
Wynndow: Could you shift this up to the `init()` to save calling it everytime?

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
```diff
@@ -0,0 +1,263 @@
...
+
+        assertThat(response.getStatusInfo()).isEqualTo(Response.Status.INTERNAL_SERVER_ERROR);
+    }
+
+    private String getEntityId(String name) {
+        return "http://localhost:" + testMetadataServer.getPort() + "/application/" + name + "/Metadata";
+    }
+
+    private String getUriString(String entityId) {
+        return "http://" + UriBuilder.fromMethod(MetatronResource.class, "metadata").host("localhost").port(metatronAppRule.getLocalPort()).build(entityId).toString();
```
Wynndow: This is interesting - looks like it URL encodes the entityId automatically. Might be an update to the client I merged today.

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetatronConfiguration.java
Wynndow: The Prometheus bundle is still being loaded in the Application class?

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
JohnWattsGDS: As it turns out, for reasons I don't understand (but related to how class rules and static things come to life I guess) - I couldn't!

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
JohnWattsGDS: I suspect it's not limited to the .fromMethod() method would probably work a treat.

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasConfig.java
JohnWattsGDS: I thought I'd replied to this earlier but it seems not.  I take it as being an annotation in the simplest sense of the word - it just tells the casual observer that this is going to be populated by some json.  It will work without it but I feel like it's useful information to have in place.

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetatronConfiguration.java
JohnWattsGDS: It's going now that the Prometheus bundle is going (it's a config requirement of the PrometheusBundle)

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
JohnWattsGDS: I changed my mind about this and went with the Optional getter.  It _is_ better to signal that there may not be a TLS Truststore and actually, as long as it's just the getter that returns the optional, my concerns about the jackson side being more complicated are nonsense.

Wynndow:
Codacy upset about a bunch of unused imports

CharlesIC:
Yep we shouldn't have unused imports

CharlesIC:
Also the second commit with minor fixed can be squashed into the first one

JohnWattsGDS:
The second commit is a courtesy to Chris so that he can see what I’ve done in response to his comments and will be squashed once everyone has had their say before approving.

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetatronApplication.java
```diff
@@ -31,7 +37,7 @@ public static void main(final String[] args) throws Exception {
 
     @Override
     public String getName() {
-        return "Metatron - service in disguise";
+        return "Metatron - Provides all of your CountryMetadataResponse needs. Does not provide MetatronResponses";
```
Wynndow: 🤭

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasCountryConfig.java
```diff
@@ -62,6 +70,22 @@ public void setEnabled(boolean enabled) {
         this.enabled = enabled;
     }
 
+    public void setMetadataTruststore(KeyStore metadataTruststore) {
```
Wynndow: The setters in here are never used, and probably never will be as the values are being deserialized from the country config. Do we need them at all?

metatron/src/main/java/uk/gov/ida/eidas/metatron/MetatronApplication.java
CharlesIC: 😂
I approve this message

JohnWattsGDS:
Since I’ll need to squash the second commit anyway, I may as well remove them. It would signpost that the content only comes from a config file so is worth it.

metatron/build.gradle
```diff
@@ -4,8 +4,11 @@ version '1.0-SPIKE'
 dependencies {
     compile configurations.dropwizard,
```
idrop:
```suggestion
    implementation configurations.dropwizard,
```

metatron/build.gradle
```diff
@@ -4,8 +4,11 @@ version '1.0-SPIKE'
...
             configurations.common,
-            'com.sun.xml.bind:jaxb-core:2.3.0',
-            'com.sun.xml.bind:jaxb-impl:2.3.0'
+            configurations.verify_saml,
+            configurations.opensaml,
+            project(':proxy-node-shared')
+
+    testCompile 'com.github.stefanbirkner:system-rules:1.19.0'
```
idrop:
```suggestion
    testImplementation 'com.github.stefanbirkner:system-rules:1.19.0'
```

metatron/build.gradle
idrop: There's a newer way of doing this that aligns with a recent PR that landed

JohnWattsGDS: Can we leave this/these until the next iteration?  Chris will have to approve it a third time and Karol will have a fit about the extra commits that made it into the PR if I use the GitHub Commit suggestion button.

idrop: 👍 okay

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/EidasCountryConfig.java
```diff
@@ -19,47 +23,39 @@
...
-    public boolean isEnabled() {
-        return enabled;
+    public KeyStore getMetadataTruststore() {
+        return this.metadataTruststore;
     }
 
-    public void setEnabled(boolean enabled) {
-        this.enabled = enabled;
+    public Optional<KeyStore> getTlsTruststore() {
+        return Objects.nonNull(this.tlsTruststore) ? Optional.of(this.tlsTruststore) : Optional.empty();
```
idrop:
```suggestion
        return Optional.ofNullable(this.tlsTruststore);
```

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public MetadataResolver getMetadataResolver(String entityId) {
+        return metadataResolverMap.get(entityId);
+    }
+
+    private Client getClient(EidasCountryConfig country) {
+        return country.getTlsTruststore().isPresent() ?
+                JerseyClientBuilder.newBuilder().trustStore(country.getTlsTruststore().get()).build()
+                :
```
jkunle: Very minor with the  ternary operator should be symmetric i.e : at the end
instead of a separate line.
```
return country.getTlsTruststore().isPresent()
                ? JerseyClientBuilder.newBuilder().trustStore(country.getTlsTruststore().get()).build() 
                : JerseyClientBuilder.newClient();
```

metatron/src/test/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverServiceTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void shouldProvideCorrectMetadataResolverForEntityId() {
+
+        KeyStore keyStore = new KeyStoreLoader().load(ResourceHelpers.resourceFilePath("test-truststore.ts"), "puppet");
+        when(countriesConfig.getCountries()).thenReturn(List.of(countryConfig1, countryConfig2));
+        setCountryConfig(countryConfig1, keyStore, AN_ENTITY_ID, metadataResolver1);
+        setCountryConfig(countryConfig2, keyStore, "http://localhost/another-entity", metadataResolver2);
+        when(countryConfig2.getTlsTruststore()).thenReturn(Optional.of(keyStore));
+        when(metadataResolverFactory.create(any(Client.class), eq(URI.create(AN_ENTITY_ID)), any(List.class), eq(MIN_REFRESH_DELAY_MS), eq(MAX_REFRESH_DELAY_MS))).thenReturn(metadataResolver1);
```
Wynndow: I think this is already being achieved by the call to `setCountryConfig` on line 53?

PR #456, opened 2020-01-31
EID-1828 Create Metatron API endpoint

idrop:
https://govukverify.atlassian.net/browse/EID-1828

Provide a REST API which will be called from the `esp` and `translator` services.
Given a country's connector node entity id, the API provides data about that country and its metadata.

Unit tests are dropped in favour of testing real cases with Dropwizard AppRule tests.

The in-memory map of `country` -> `metadata resolver` in MetadataResolver service now uses `EidasCountryConfig` as key, so country info can be stored in the key, like the country code.

The API returns either a:
* 200 with a json representation of `CountryMetadataResponse`
* 400 with a json error map when the client has sent a bad request
* 500 with a json error map when the metatron has a processing error


metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -58,5 +93,74 @@ private MetadataResolver createMetadataResolver(EidasCountryConfig country) {
...
+        }
+        return eidasCountryConfig;
+    }
+
+    private MetadataResolver getMetadataResolver(EidasCountryConfig eidasCountryConfig) {
+        return metadataResolverMap.get(eidasCountryConfig);
+    }
+
+    private String getCertificateAsX509(EidasCountryConfig countryConfig, UsageType usageType) {
+        X509Credential credential = (X509Credential) getCredential(usageType, countryConfig, SPSSODescriptor.DEFAULT_ELEMENT_NAME);
```
Wynndow: As the `DEFAULT_ELEMENT_NAME` is an imported constant you could use it directly in `getCredential` rather than have it as a param.

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -1,39 +1,74 @@
...
+                .collect(Collectors.toMap(Function.identity(), this::createMetadataResolver));
     }
 
-    public MetadataResolver getMetadataResolver(String entityId) {
-        return metadataResolverMap.get(entityId);
+    public CountryMetadataResponse getCountryMetadataResponse(URI entityId) {
+        EidasCountryConfig countryConfig = getCountryConfig(entityId);
+        URI location = getAssertionConsumerServiceLocation(countryConfig);
+        String encryptionX509 = getCertificateAsX509(countryConfig, UsageType.ENCRYPTION);
+        String signingX509 = getCertificateAsX509(countryConfig, UsageType.SIGNING);
```
Wynndow: These variable assignments could be dropped and the method calls inlined into the constructor.

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -1,39 +1,74 @@
...
+            CredentialResolverFactory credentialResolverFactory) {
         this.metadataResolverFactory = metadataResolverFactory;
-        this.metadataResolverMap = new HashMap<>();
+        this.credentialResolverFactory = credentialResolverFactory;
         this.expiredCertificateMetadataFilter = new ExpiredCertificateMetadataFilter();
-
-        metadataResolverMap = countriesConfig.getCountries().stream().collect(
-                Collectors.toMap(EidasCountryConfig::getEntityId, this::createMetadataResolver));
+        this.metadataResolverMap = countriesConfig.getCountries().stream()
+                .collect(Collectors.toMap(Function.identity(), this::createMetadataResolver));
```
Wynndow: What's the benefit of having the key be the actual config, rather than the entity id? It seems it means implementing the logic to initially get the config out of the map by iterating through all the keys and checking against a given entity id. And once we've done that once we use that config twice more to access the map again to get the metadata resolvers.
It seems like it might be easier to just initially access it via the entity id, and then just call `getEntityId()` on that config when we need to fetch the metadata from the map in future.

metatron/src/main/java/uk/gov/ida/eidas/metatron/exceptions/MetatronClientException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.ida.eidas.metatron.exceptions;
+
+public class MetatronClientException extends RuntimeException {
```
Wynndow: I find the naming of the exception a bit odd. If I saw it I'd think it came from the client we use to call Metatron from the other apps. Where we're throwing it, it's not really an issue with the client, more with the request that the client made.

Having said that we call the client `MetatronProxy` so maybe it's ok. Not sure why we call it `MetatronProxy` and not `MetatronClient` though....

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
```diff
@@ -82,7 +85,12 @@ public void firstValidResolvesOk() {
         Response response = getClient().target(getUriString(entityId)).request().get();
 
         assertThat(response.getStatusInfo()).isEqualTo(Response.Status.OK);
-        assertThat(response.readEntity(String.class)).isEqualTo(entityId);
+        CountryMetadataResponse actual = response.readEntity(CountryMetadataResponse.class);
```
Wynndow: Could this be moved up to under the call to `getClient()`, so all the assert are grouped at the end?

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
```diff
@@ -91,7 +99,12 @@ public void secondValidResolvesOk() {
...
```
Wynndow: As above.

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
JohnWattsGDS: I'm with Chris on this one - it seems like over-egging the pudding a bit.  If I've understood this correctly, this is a way of getting the country code and the entityId to put into the CountryMetadataResponse but I really feel that the keys should have a single responsibility of being a key; it might be a clearer statement of intent to have an object composed of the properties we want and the metadata resolver as value; we'd retain the advantages of a map that way too.  I have crazy ideas about CountryMetadataResponses that contain Functions and Suppliers that get called by the getters but I've just said we should probably keep it simple so I won't suggest them!

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
```diff
@@ -136,13 +173,26 @@ private Client getClient() {
         return client;
     }
 
+    private void assertThatCertsMatch(String samlEncryptionCertX509, String testRpPublicEncryptionCert) {
+        assertThat(StringUtils.right(samlEncryptionCertX509, 10)).isEqualTo(StringUtils.right(testRpPublicEncryptionCert, 10));
```
JohnWattsGDS: Is it not possible to compare the entire string?  These sort of comparison tests always make me nervous because we're never sure they're truly accurate.

Would you consider turning both the strings into certs and doing `.isEqualToComparingFieldByField()` if the whole strings can't be compared for some reason?

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -58,5 +93,74 @@ private MetadataResolver createMetadataResolver(EidasCountryConfig country) {
                 MAX_REFRESH_DELAY_MS);
         return resolver;
     }
-}
 
+    private EidasCountryConfig getCountryConfig(URI entityId) {
```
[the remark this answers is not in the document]
idrop: Sure, updated

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -58,5 +93,74 @@ private MetadataResolver createMetadataResolver(EidasCountryConfig country) {
...
+                .filter(c -> entityId.equals(c.getEntityId()))
+                .findFirst()
+                .orElseThrow(() -> new MetatronClientException(String.format("No metadata entry for entityID %s", entityId)));
+        if (!eidasCountryConfig.isEnabled()) {
+            throw new MetatronClientException(String.format("EntityID %s is not enabled", entityId));
+        }
+        return eidasCountryConfig;
+    }
+
+    private MetadataResolver getMetadataResolver(EidasCountryConfig eidasCountryConfig) {
```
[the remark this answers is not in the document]
idrop: Have done

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
idrop: Yes, done

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
idrop: The entityId changes per request to this method, so they can't be inlined to constructor

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
idrop: Addressed.

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
idrop: Yes, done

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
idrop: 👍, tests now compare X509Certificates for equality

metatron/src/test/java/uk/gov/ida/eidas/metatron/apprule/rules/MetatronTests.java
idrop: Done.

PR #462, opened 2020-02-03
Try to fix parallel stuff in pipeline

Krenair:
Redoes reverts commit 920530e7634355f32a8536d8c85ba38bad07b7d2 in a different
way.

Continue doing things in parallel, but be careful about build steps being done
before we put the result to the image repository. Also ensure that the output
directory names are unique to avoid one image being built and pushed to a
different image's repository.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -257,7 +257,6 @@ spec:
       - get: cloudhsm-config
         trigger: true
       - in_parallel:
-          limit: 3
           fail_fast: true
           steps:
           - do:
@@ -348,7 +347,6 @@ spec:
                 ./gradlew --console verbose --parallel -Pcloudhsm -PCI installDist
 
       - in_parallel:
-          limit: 3
           fail_fast: true
           steps:
           - do:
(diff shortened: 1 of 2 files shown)


ci/build/build-pipeline.yaml
```diff
@@ -256,31 +256,43 @@ spec:
...
-          params:
-            CONTEXT: cloudhsm-config/cloudhsm/jdk-jce-image
-          inputs:
-          - name: cloudhsm-config
-      - put: cloudhsm-jce-image
-        params:
-          <<: *image_put_params
-          additional_tags: cloudhsm-config/.git/short_ref
+      - in_parallel:
+          limit: 3
```
chrisfarms: You can probably remove these limits now... It was only there previous to reduce push errors to harbor (and there's only two do blocks)

ci/build/build-pipeline.yaml
```diff
@@ -256,31 +256,43 @@ spec:
...
+          fail_fast: true
+          steps:
+          - do:
+            - task: build-cloudhsm
+              privileged: true
+              config:
+                <<: *build_config
```
chrisfarms: I think use of this anchor is bit confusing at this point... maybe the anchor should just be for the `image_resource` part (since everything else params/inputs/outputs appears to be overwritten in each use of the task) so it's a bit misleading and looks like some magic merging is going on

PR #465, opened 2020-02-04
no rootfs param for alpine-image with registry-image resource

Krenair:
format parameter defaults to rootfs


ci/build/build-pipeline.yaml
```diff
@@ -490,7 +490,6 @@ spec:
           - get: cloudhsm-client-image
             passed: [build-cloudhsm]
           - get: alpine-image
-            params: {rootfs: true}
```
blairboy362: params: {format: oci}

PR #466, opened 2020-02-04
Replace alpine-image hackery

Krenair:
We've spent enough time trying to make this hack work in the new world.
Make it a bit less hacky.

### DIFF
--- a/ci/sandbox/build-pipeline.yaml
+++ b/ci/sandbox/build-pipeline.yaml
@@ -102,13 +102,6 @@ spec:
         username: "re-autom8-ci"
         password: ((github.api-token))
 
-    - name: alpine-image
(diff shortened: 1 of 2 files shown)


ci/sandbox/build-pipeline.yaml
```diff
@@ -598,11 +587,36 @@ spec:
...
+        outputs:
+        - name: chart-image-file
+          path: image
+        run:
+          path: /bin/ash
+          args:
+          - -euc
+          - |
+            cat <<<EOF > ./Dockerfile
+            FROM alpine:3
```
chrisfarms: `FROM scratch` maybe

https://hub.docker.com/_/scratch

PR #478, opened 2020-02-10
BAU slack on build pipeline job fail

idrop:
Send a slack message to `#verify-eidas-alerts` when a concourse job fails on the Proxy Node [build](https://ci.london.verify.govsvc.uk/teams/eidas-proxy-node-build/pipelines/build-release) pipeline.

The slack webhook url required by the resource is referenced in the pipeline via the `((slack-notification.url))` var. The value of this url is set via a sealed secret targeting the namespace.

Secrets are sealed [using this method](https://verify-team-manual.cloudapps.digital/documentation/support/eidas-proxy-node/connecting-vsp-to-hub.html#creating-a-sealed-secret-for-the-vsp).

The interpolation of yaml anchors into aliases can be validated by [pasting in the pipeline yaml here](https://yaml-online-parser.appspot.com/) and checking the `on_failure` node is set correctly.


Wynndow (approved):
Looks good. Although I am mostly just ignoring that slack channel now as it's so noisy. We should look into fixing all those alerts.

JohnWattsGDS (approved):
Like Chris, I don't really notice the notifications any more.  Brendan and I had a crack at ~grassing~ notifying the last committer at some point in the past but it never made it out of our test-bed. (https://github.com/alphagov/verify-infrastructure-config/blob/7c6665c8a37ce5f8175270413806463ecce52e8d/pipelines/builder/build-libraries.yml#L216).  Maybe we can discuss that for a future iteration to increase the signal to noise ratio?

PR #480, opened 2020-02-10
EID-1835 Use MetatronClient in EidasSamlApplication (ESP)

JohnWattsGDS:
Remove MetadataResolverBundle and related classes.

Add required MetatronProxy config values and inject MetatronProxy where required.

Use MetatronProxy in EidasSamleResource and move code into methods representing logical units.

Update tests to work with MetatronProxy - in particular create TestMetatronResource so ClientRules can return CountryMetadataResponses.  Remove test that expects a healthcheck result from MetadataResolver.


eidas-saml-parser/src/dist/config.yml
```diff
@@ -16,17 +16,12 @@ logging:
...
-  expectedEntityId: ${CONNECTOR_NODE_ENTITY_ID:-DefaultNotUsed}
-  jerseyClientName: connector-metadata-client
-  maxRefreshDelay: ${METADATA_REFRESH_DELAY:-600000}
-  trustStore:
-    type: ${TRUSTSTORE_TYPES:-encoded}
-    store: ${CONNECTOR_NODE_METADATA_TRUSTSTORE:-DefaultNotUsed}
-    password: ${CONNECTOR_NODE_METADATA_TRUSTSTORE_PASSWORD:-marshmallow}
+metatronUri: ${METATRON_URL:-https://DefaultNotUsed}
+
+url: http://foo.com
```
JohnWattsGDS: I don't really know what this does ...

Wynndow: Looks like it's not used anywhere. I think you can drop it, as well as the getter in `EidasSamlParserConfiguration`

eidas-saml-parser/src/dist/config.yml
```diff
@@ -16,17 +16,12 @@ logging:
...
-connectorMetadataConfiguration:
-  url: ${CONNECTOR_NODE_METADATA_URL:-https://DefaultNotUsed}
-  expectedEntityId: ${CONNECTOR_NODE_ENTITY_ID:-DefaultNotUsed}
-  jerseyClientName: connector-metadata-client
-  maxRefreshDelay: ${METADATA_REFRESH_DELAY:-600000}
-  trustStore:
-    type: ${TRUSTSTORE_TYPES:-encoded}
-    store: ${CONNECTOR_NODE_METADATA_TRUSTSTORE:-DefaultNotUsed}
-    password: ${CONNECTOR_NODE_METADATA_TRUSTSTORE_PASSWORD:-marshmallow}
+metatronUri: ${METATRON_URL:-https://DefaultNotUsed}
```
Wynndow: Maybe drop the default? It's always going to be required I think.

JohnWattsGDS: I figured that meant that there is no default/don't use a default judging by other places it shows up.  Delete the other `DefaultNotUsed`s?

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
```diff
@@ -29,44 +39,55 @@
...
+        return new EidasSamlParserResponse(
+                authnRequest.getID(),
+                authnRequest.getIssuer().getValue(),
+                metatronResponse.getSamlEncryptionCertX509(),
+                metatronResponse.getDestination().toString());
+    }
+
+    private AuthnRequest unmarshallRequest(EidasSamlParserRequest request) throws UnmarshallingException, XMLParserException {
+        String xml = new String(Base64.getDecoder().decode(request.getAuthnRequest().getBytes()));
+        System.out.println(xml);
```
Wynndow: Should probably drop these two lines

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
```diff
@@ -75,4 +96,34 @@ private String getLevelsOfAssurance(AuthnRequest authnRequest) {
                 .map(AuthnContextClassRef::getAuthnContextClassRef)
                 .collect(Collectors.joining(","));
     }
+
+    private Credential getSigningCredential(String base64X509Certificate) throws CertificateException {
+        CertificateFactory certificateFactory = CertificateFactory.getInstance("X509");
+
+        byte[] decodedCert = Base64.getMimeDecoder().decode(base64X509Certificate);
```
Wynndow: Does this need to be a MIME decoder? I had to look it up. Looks like B64 with added formatting: https://www.tutorialspoint.com/java8/java8_base64.htm

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
```diff
@@ -75,4 +96,34 @@ private String getLevelsOfAssurance(AuthnRequest authnRequest) {
...
+        return signingCredential;
+    }
+
+    private void validateIssuer(AuthnRequest authnRequest) {
+        if (authnRequest.getIssuer().getValue() == null) {
+            throw new SamlTransformationErrorException("Issuer is missing", Level.WARN);
+        }
+    }
+
+    private void isRequestSignatureValid(AuthnRequest authnRequest, Credential signingCredential) {
```
Wynndow: Is it worth moving the logic about signatures and credential retrieval etc to some sort of helper or validator? A bit like the AssertionConsumerServiceValidator below maybe?

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
JohnWattsGDS: If you change it to a standard decoder lots of tests fail.  The MimeDecoder is the more robust approach and I'm sure it's used elsewhere in the Verify codebase (as that's where I first saw it) - we could alter whatever test data it finds offensive but by using it, we build in some robustness and keep [Jon Postel](https://en.wikipedia.org/wiki/Robustness_principle) happy.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserConfiguration.java
```diff
@@ -19,19 +27,36 @@
...
 
+    @NotNull
     @JsonProperty
+    private URI url;
+
     @Valid
     @NotNull
-    private TrustStoreBackedMetadataConfiguration connectorMetadataConfiguration;
+    @JsonProperty
+    private final JerseyClientConfiguration clientConfig = new JerseyClientConfiguration();
```
Wynndow: Is the reason to have this as a field, rather than just instantiated in `buildMetatronProxy()` down below so that we have a default but can also override it if needed? If so, I think that `config.yml` for the ESP will need an update?

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/saml/validation/components/AssertionConsumerServiceValidator.java
```diff
@@ -1,71 +1,33 @@
...
+    public AssertionConsumerServiceValidator(MetatronProxy metatronProxy) {
+        this.metatronProxy = metatronProxy;
     }
 
     public void validate(AuthnRequest request) {
         String assertionConsumerServiceURL = request.getAssertionConsumerServiceURL();
-
         if (Strings.isNullOrEmpty(assertionConsumerServiceURL)) {
             return;
         }
```
idrop: if a request has no assertionConsumerServiceURL we should throw a 400

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserConfiguration.java
Wynndow: Otherwise I think `buildMetatronProxy()` could be defined somewhere else, like the Application class rather than in here, as all it really needs is the enviroment.

eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/base/TestMetatronResource.java
```diff
@@ -0,0 +1,29 @@
...
+import java.net.URI;
+
+import static uk.gov.ida.saml.core.test.TestCertificateStrings.TEST_RP_PUBLIC_ENCRYPTION_CERT;
+import static uk.gov.ida.saml.core.test.TestCertificateStrings.TEST_RP_PUBLIC_SIGNING_CERT;
+
+@Path("/")
+public class TestMetatronResource {
+
+    @GET
+    @Path("/{entity-id}")
```
Wynndow: Should the path be `"metadata/{entity-id}"? That's what got merged recently or are these tests set up different?

eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/EidasSamlResourceTest.java
```diff
@@ -57,17 +61,31 @@ public void shouldReturnRequestIdAndIssuer() throws Exception {
...
                 .build();
-        setLevelOfAssurance(authnRequest, EidasLoaEnum.LOA_SUBSTANTIAL);
-        EidasSamlParserRequest request = new EidasSamlParserRequest(Base64.encodeAsString(ObjectUtils.toString(authnRequest)));
+
+        EidasSamlParserRequest request = new EidasSamlParserRequest(Base64.encodeAsString(new SamlObjectMarshaller().transformToString(eidasAuthnRequest)));
+
+        CountryMetadataResponse metatronResponse =
+                new CountryMetadataResponse(
+                    METADATA_SIGNING_A_PUBLIC_CERT,
+                        TEST_RP_PUBLIC_ENCRYPTION_CERT,
```
Wynndow:
```suggestion
                    TEST_RP_PUBLIC_ENCRYPTION_CERT,
```

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/saml/validation/components/AssertionConsumerServiceValidator.java
JohnWattsGDS: Makes sense - I used the existing logic so good spot!

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
```diff
@@ -29,44 +39,55 @@
...
+        String xml = new String(Base64.getDecoder().decode(request.getAuthnRequest().getBytes()));
+        System.out.println(xml);
+        return ObjectUtils.unmarshall(
+                new ByteArrayInputStream(xml.getBytes()),
+                AuthnRequest.class);
+    }
+
+    private CountryMetadataResponse getMetatronResponse(AuthnRequest authnRequest) {
+        this.validateIssuer(authnRequest);
+        return metatronProxy.getCountryMetadata(authnRequest.getIssuer().getValue());
```
idrop: after this line we should check that authnRequest issuer matches CountryMetadataResponse.entityid

metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -141,7 +141,7 @@ private URI getAssertionConsumerServiceLocation(CountryMetadata countryMetadata)
             throw new MetatronServerException(String.format("Unable to resolve entityDescriptor with given criteria for entityId %s", entityId), e);
         }
         if (entityDescriptor == null) {
-            throw new MetatronServerException(String.format("No entityDescriptor for entityId %s"));
+            throw new MetatronServerException(String.format("No entityDescriptor for entityId %s", entityId));
```
idrop: oops 😄

proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/MetatronSamlSignatureValidator.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.ida.notification.saml;
+
+public class MetatronSamlSignatureValidator {
```
idrop: Beautifully named, no methods, no-side effects.

JohnWattsGDS: 100% test coverage too ...

eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/IstioHeaderTest.java
```diff
@@ -31,6 +32,9 @@
     private static final DropwizardClientRule metadataClientRule = createTestMetadataClientRule();
     private static final EidasSamlParserAppRule eidasSamlParserAppRule = createEidasSamlParserRule(metadataClientRule);
 
+    @ClassRule
+    public static final DropwizardClientRule metatronService = createInitialisedClientRule(new TestMetatronResource());
```
Wynndow: Is this actually being used? I'm trying to run the test but intellij is unhappy.

proxy-node-shared/src/main/java/uk/gov/ida/notification/saml/MetatronSamlSignatureValidator.java
```diff
@@ -0,0 +1,5 @@
...
+
```
Wynndow: Could put the signature stuff in here!

eidas-saml-parser/src/dist/config.yml
Wynndow: hmm..interesting. Not sure in that if that's the case. Seems weird to supply a default if the intention is to stop you using one. Surely _not_ providing one would force you to supply a value.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java
Wynndow: How is it more robust?

JohnWattsGDS: If you change it to Base64.getDecoder()... you'll see this exception thrown in lots of tests:

```! java.lang.IllegalArgumentException: Illegal base64 character a```

The only thing that has changed is the decoder.  It might be that we'll never come across something with the offending encoding in practice but if we do, it will work.  For the record, I'm not sure why a would be an illegal character but the MimeDecoder handles input that the vanilla one doesn't.

File: eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java (hunk 5 above)
Comment (JohnWattsGDS) [reply eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java]: Credentials I think we should hold off on until we've considered any changes we might make to the `CountryMetadataResponse` (i.e. whether we decode the base64 when it comes in and present an actual certificate in the DTO - maybe we also present the credentials or present those instead?).

Signature validation; as discussed in the office, there's a better candidate for that.

File: eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/base/TestMetatronResource.java (hunk 8 above)
Comment (JohnWattsGDS) [reply eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/base/TestMetatronResource.java]: The tests are set up to use the uri of the test resource so that's why it works anyway but I've changed it so that it documents things (like Metatron's endpoint) better.

File: eidas-saml-parser/src/dist/config.yml (hunk 2 above)
Comment (JohnWattsGDS) [reply eidas-saml-parser/src/dist/config.yml]: I tend to agree and I've deleted defaultNotUsed wherever I found it.

File: eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserConfiguration.java (hunk 6 above)
Comment (JohnWattsGDS) [reply eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserConfiguration.java]: It does need the configuration too but that can be supplied in the Application class so I've moved it.  We do need the metatronUri as a config value somehow.

File: eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java (hunk 10 above)
Comment (idrop) [reply eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlResource.java]: ☝️ John and I talked this through, and decided this check would add no extra value.

**Comment by JohnWattsGDS** [comment]
Looking for a re-review @Wynndow @idrop 

I'm not too worried about the codacy warning at this stage.  It's complaining that the validate method is complicated. It is but it's had an extra if statement added and it isn't even nested so it's virtually the same as it was.

File: eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserConfiguration.java [hunk 15]
```diff
@@ -19,19 +19,28 @@
 
     @JsonProperty
     @Valid
-    private ReplayCheckerConfiguration replayChecker = new ReplayCheckerConfiguration();
+    private final ReplayCheckerConfiguration replayChecker = new ReplayCheckerConfiguration();
 
+    @Valid
+    @NotNull
     @JsonProperty
+    private final JerseyClientConfiguration clientConfig = new JerseyClientConfiguration();
```
Wynndow: I think that the `clientConfig` field can be dropped, as well as the getter for it below. The only place it would be used is in the construction of the Metatron proxy, and we don't use it there.

Wynndow: Alternatively I guess we should pull this configuration in, so that's it's customisable via the config but has a default?

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/saml/validation/EidasAuthnRequestValidator.java
```diff
@@ -49,11 +53,21 @@ public EidasAuthnRequestValidator(RequestIssuerValidator requestIssuerValidator,
...
     }
 
-    public void validate(AuthnRequest request) {
+    public void validate(AuthnRequest request, Credential signingCredential) {
+
         if (request == null) {
             throw new InvalidAuthnRequestException("Null request");
         }
 
+        try {
```
Wynndow: Nice.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserConfiguration.java
JohnWattsGDS: I've gone with taking it out.  If it turns out we do need it, we can put it back in but less is more where possible.

PR #485, opened 2020-02-11
Remove reliance on GSP-provided key for signing artefacts.

blairboy362:
The secret containing the GPG key for the cluster that is provided by
the GSP will be removed soon. Re-instate it in a tenant-controlled way.


ci/build/artefact-signing-key-sealed-secret.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: bitnami.com/v1alpha1
+kind: SealedSecret
+metadata:
+  creationTimestamp: null
+  name: artefact-signing-key
+  namespace: verify-eidas-proxy-node-build
```
Krenair: For master this would be correct but on release/single-country-legacy I think this would be:
```suggestion
  namespace: verify-proxy-node-build
```

ci/build/artefact-signing-key-sealed-secret.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  template:
+    metadata:
+      name: artefact-signing-key
+      namespace: verify-eidas-proxy-node-build
```
Krenair:
```suggestion
      namespace: verify-proxy-node-build
```

ci/build/artefact-signing-key-sealed-secret.yaml
blairboy362: Updated.

ci/build/artefact-signing-key-sealed-secret.yaml
blairboy362: I've removed this whole `template` block as I'm not convinced it's needed/adding anything useful.

PR #497, opened 2020-02-19
EID-1836: Use Metatron proxy and values in translator

Wynndow:
This PR uses the shared Metatron proxy to fetch values from the countries
metadata and use the encryption certificate and country code when
generating an eIDAS response.

There is a slight change to the gateways session object to allow the
entity ID to be passed through to the translator - this is necessary so
that the translator can call Metatron correctly.

As the translator is now fetching the metadata for each requests we can
remove the hard coded values for each country from the configuration
class and yaml.

Co-authored-by: Olakunle Jegede <olakunle.jegede@digital.cabinet-office.gov.uk>
Co-authored-by: Phil Miller <phillip.miller@digital.cabinet-office.gov.uk>


proxy-node-gateway/src/test/java/uk/gov/ida/notification/apprule/HubResponseAppRuleTests.java
```diff
@@ -155,7 +156,7 @@ public void redisCanStoreCertificateInSession() throws Throwable {
         assertThat(response.getStatus()).isEqualTo(200);
 
         final List<HubResponseTranslatorRequest> translatorArgs = TEST_TRANSLATOR_RESOURCE.getTranslatorArgs();
-        assertThat(translatorArgs.get(0).getConnectorEncryptionCertificate()).isEqualTo(STUB_COUNTRY_PUBLIC_PRIMARY_CERT);
+        assertThat(translatorArgs.get(0).getEidasIssuerEntityId()).isEqualTo(URI.create(SAMPLE_ENTITY_ID));
```
JohnWattsGDS: Should the name of the test be changed as well?

Wynndow: yep

PR #499, opened 2020-02-19
BAU: See new host pattern for HMRC

Wynndow:
Use conditional logic to use the new host pattern for HMRC


CharlesIC:
This is no longer needed as RE have added the old-style URLs to the DNS whitelist

PR #500, opened 2020-02-19
EID-1837 Clean up DTOs used between Gateway and ESP

JohnWattsGDS:
We no longer need to pass the encryption certificate around as it's possible to get the information from Metatron.

Remove connectorEncryptionPublicCertificate from EidasSamlParserResponse.

Update any use of EidasSamlParserResponse so that code will compile.

Update tests so that they reflect new DTO.

*NOTE* this depends on eid-1836 as it's branched from there - there's some dependency between because of the Redis change in eid-1836

### DIFF
--- a/proxy-node-shared/src/main/java/uk/gov/ida/notification/session/storage/RedisStorageService.java
+++ b/proxy-node-shared/src/main/java/uk/gov/ida/notification/session/storage/RedisStorageService.java
@@ -42,7 +42,7 @@ public class RedisStorageService extends AbstractStorageService {
     }
 
     @Override
-    public StorageRecord read(String context, String key) throws IOException {
+    public StorageRecord read(String context, String key) {
         List<KeyValue<String, String>> result = redis.hmget(key(context, key), "value", "version", "expiration");
         String value = result.get(0).getValueOrElse(null);
         if (value == null) return null;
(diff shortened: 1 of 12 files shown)


proxy-node-shared/src/main/java/uk/gov/ida/notification/session/storage/RedisStorageService.java
```diff
@@ -99,7 +99,7 @@ public Long updateWithVersion(long version, String context, String key, String v
             return null;
         } else if (storedVersion != version) {
             redis.unwatch();
-            throw new VersionMismatchException(String.format("Expected version to be %l but it was %l", version, storedVersion));
+            throw new VersionMismatchException(String.format("Expected version to be %d but it was %d", version, storedVersion));
```
Wynndow: I can't find a reference to the "%l" format specifier in Java anywhere. In C it's for a long. Was this just a mistake?

JohnWattsGDS: I ran the IntelliJ analyse thing and it spotted it - I think whoever did it, guessed it was *long* (maybe from C experience) but that as you say, isn't a thing in Java.  I'm pretty sure I've done it myself in the past but spotted it because I ran the code.  All that would happen is that if that exception was thrown it would output the literal string with `%l` in it instead of the substituted variable.  I would be surprised if that exception has ever been thrown so nobody (except IntelliJ) has ever noticed.

Wynndow:
Any chance of dropping the last 6 commits?

JohnWattsGDS:
What commits? 😉

Wynndow:
Oh weird. I must've imagined it...

JohnWattsGDS:
History is mutable

PR #503, opened 2020-02-21
BAU Un-break the broken build

JohnWattsGDS:
Some time ago, builds stopped building.  The test that fails doesn't fail locally or when Travis runs it.  This suggests some kind of race is going on and the wrong thing wins in the pipeline.  It looks like it's the mockatron that's providing the null pointer exceptions when it should be providing port numbers so ~I've changed the order in which~ make the test wait for the port to become available before the RuleChain is executed to see if that helps.


CharlesIC:
This is fine but should have gone into the base class `AbstractSamlAppRuleTestBase` as we might need this for other tests too

PR #505, opened 2020-02-21
🍒BAU: Preserve permanent Sanbox branch config

CharlesIC:
Cherry pick from the Sandbox branch so the permanent config needed for this branch doesn't get accidentally deleted when rebasing on master


CharlesIC:
This was done in #510

PR #506, opened 2020-02-21
BAU Back to Basics for the Broken Build

JohnWattsGDS:
Use the @ClassRule annotation (as the designers of Dropwizard intended) to hopefully ensure we have everything we want when we need it.


CharlesIC:
This doesn't fix it - another PR with a fix coming soon

PR #507, opened 2020-02-21
BAU: Fix for ESP tests and minor test refactoring

CharlesIC:
Various fixes and improvements:
- 🍒Cherry pick the `build-pipeline.yml` file from the Sandbox branch so the permanent config needed for this branch doesn't get accidentally deleted when rebasing on master

- Fix ESP tests: Remove test specific client and app rules from the base app rule test class and make each child initialise those individually instead.

- Apply the app rule tests pattern to Metatron tests
  - Refactor Metatron app rule tests to use the Dropwizard `@ClassRule`s and `RuleChain` to define dependencies between test objects

- Create a base AppRule class
We mostly copy the default implementation so this can be easily refactored into a base class.
The copies of the default implementation have been removed.

These changes have passed in Sandbox:
https://ci.london.sandbox.govsvc.uk/teams/proxy-node-dev/pipelines/build-test-sandbox/jobs/build-proxy-node/builds/71


eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/EspAuthnRequestAppRuleTest.java
```diff
@@ -41,6 +46,15 @@
...
 
+    @ClassRule
+    public static final DropwizardClientRule metatronClientRule = createInitialisedClientRule(new TestMetatronResource());
+
+    @ClassRule
+    public static final EidasSamlParserAppRule eidasSamlParserAppRule = createEidasSamlParserRule(metatronClientRule);
+
+    @ClassRule
+    public static final RuleChain orderedRules = RuleChain.outerRule(metatronClientRule).around(eidasSamlParserAppRule);
+
```
Wynndow: This felt better on the base class to me. The only configurable part of the setup is the test resource used for the Metatron stub, and we use the same thing in the two children of the base class. Keeping them on the base removes a bunch of duplication and means the `eidasSamlParserAppRule` doesn't need passing into so many calls to `postEidasAuthnRequest` unnecessarily.

stub-connector/src/test/java/uk/gov/ida/notification/apprule/base/StubConnectorAppRuleTestBase.java
```diff
@@ -21,18 +23,19 @@
...
-    public static final String ENTITY_ORG_NAME = "stub country org name";
-    public static final String ENTITY_ORG_DISPLAY_NAME = "stub country org display name";
-    public static final String ENTITY_ORG_URL = "http://stub-connector/homepage";
+
+    private static final KeyStoreResource METADATA_TRUSTSTORE = createMetadataTruststore();
 
     private Map<String, NewCookie> cookies;
 
-    protected String getEidasRequest(StubConnectorAppRule stubConnectorAppRule) throws URISyntaxException {
+    protected String getEidasRequest(AppRule<StubConnectorConfiguration> stubConnectorAppRule) throws URISyntaxException {
```
Wynndow: Again here, the only two children of `StubConnectorAppRuleTestBase` both have exactly the same setup for AppRules and ClientRules. They could be pulled back to the base class which would simplify all these methods.

eidas-saml-parser/src/test/java/uk/gov/ida/notification/eidassaml/apprule/EspAuthnRequestAppRuleTest.java
CharlesIC: Keeping those things on the base class makes tests fail - they have to instantiated by each child independently, otherwise, they get killed by Dropwizard between class runs. This is the pattern we use for other AppRule tests (e.g. Gateway). They're static so they only need to be initialised once per class, no performance issue there.

stub-connector/src/test/java/uk/gov/ida/notification/apprule/base/StubConnectorAppRuleTestBase.java
CharlesIC: Nope - that will break the tests due to how Dropwzard AppRules work in tests

PR #508, opened 2020-02-24
EID-1838: Clean up dtos between gateway and translator

jkunle:
Change DTOs between Gateway and the Translator to not pass information that can be obtained from the new Metadata Service.

### DIFF
--- a/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserApplication.java
+++ b/eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserApplication.java
@@ -3,10 +3,14 @@ package uk.gov.ida.notification.eidassaml;
 import engineering.reliability.gds.metrics.bundle.PrometheusBundle;
 import io.dropwizard.Application;
 import io.dropwizard.client.JerseyClientBuilder;
+import io.dropwizard.client.JerseyClientConfiguration;
 import io.dropwizard.configuration.EnvironmentVariableSubstitutor;
 import io.dropwizard.configuration.SubstitutingSourceProvider;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
+
+import javax.ws.rs.client.Client;
+
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
 import org.opensaml.core.config.InitializationException;
 import org.opensaml.core.config.InitializationService;
@@ -34,10 +38,8 @@ import uk.gov.ida.notification.shared.logging.ProxyNodeLoggingFilter;
 import uk.gov.ida.notification.shared.proxy.MetatronProxy;
 import uk.gov.ida.notification.shared.proxy.ProxyNodeJsonClient;
 
-import javax.ws.rs.client.Client;
-
 public class EidasSamlParserApplication extends Application<EidasSamlParserConfiguration> {
-
+    private MetatronProxy metatronProxy;
     public static void main(final String[] args) throws Exception {
         if (args == null || args.length == 0) {
             String configFile = System.getenv("CONFIG_FILE");
@@ -73,11 +75,12 @@ public class EidasSamlParserApplication extends Application<EidasSamlParserConfi
 
     @Override
     public void run(EidasSamlParserConfiguration configuration, Environment environment) throws Exception {
-
         ProxyNodeHealthCheck proxyNodeHealthCheck = new ProxyNodeHealthCheck("parser");
         environment.healthChecks().register(proxyNodeHealthCheck.getName(), proxyNodeHealthCheck);
-        MetatronProxy metatronProxy = createMetatronProxy(configuration, environment);
-        EidasAuthnRequestValidator eidasAuthnRequestValidator = createEidasAuthnRequestValidator(configuration, metatronProxy);
+
+        metatronProxy = createMetatronProxy(configuration, environment);
+
+        EidasAuthnRequestValidator eidasAuthnRequestValidator = createEidasAuthnRequestValidator(configuration);
 
         environment.jersey().register(IstioHeaderMapperFilter.class);
         environment.jersey().register(ProxyNodeLoggingFilter.class);
@@ -99,11 +102,16 @@ public class EidasSamlParserApplication extends Application<EidasSamlParserConfi
         environment.jersey().register(new CatchAllExceptionMapper());
     }
 
-    private MetatronProxy createMetatronProxy(EidasSamlParserConfiguration configuration, Environment environment) {
-        return this.buildMetatronProxy(configuration, environment);
+    private void registerInjections(Environment environment) {
+        environment.jersey().register(new AbstractBinder() {
+            @Override
+            protected void configure() {
+                bindAsContract(IstioHeaderStorage.class);
+            }
+        });
     }
 
-    private EidasAuthnRequestValidator createEidasAuthnRequestValidator(EidasSamlParserConfiguration configuration, MetatronProxy metatronProxy) throws Exception {
+    private EidasAuthnRequestValidator createEidasAuthnRequestValidator(EidasSamlParserConfiguration configuration) throws Exception {
         MessageReplayChecker replayChecker = configuration.getReplayChecker().createMessageReplayChecker("eidas-saml-parser");
         DestinationValidator destinationValidator = new DestinationValidator(
                 configuration.getProxyNodeAuthnRequestUrl(), configuration.getProxyNodeAuthnRequestUrl().getPath());
@@ -120,22 +128,15 @@ public class EidasSamlParserApplication extends Application<EidasSamlParserConfi
         );
     }
 
-    private void registerInjections(Environment environment) {
-        environment.jersey().register(new AbstractBinder() {
-            @Override
-            protected void configure() {
-                bindAsContract(IstioHeaderStorage.class);
-            }
-        });
-    }
-
-    private MetatronProxy buildMetatronProxy(EidasSamlParserConfiguration configuration, Environment environment) {
-        Client client = new JerseyClientBuilder(environment).using(environment).build("metatron-client");
-        ProxyNodeJsonClient jsonClient = new ProxyNodeJsonClient(
-                new ErrorHandlingClient(client),
-                new JsonResponseProcessor(environment.getObjectMapper()),
-                new IstioHeaderStorage()
+    private MetatronProxy createMetatronProxy(EidasSamlParserConfiguration configuration, Environment environment) {
+        JerseyClientConfiguration jerseyConfig = new JerseyClientConfiguration();
+        Client client = new JerseyClientBuilder(environment).using(jerseyConfig).build("metatron-client");
+        return new MetatronProxy(configuration.getMetatronUri(),
+            new ProxyNodeJsonClient(
+                    new ErrorHandlingClient(client),
+                    new JsonResponseProcessor(environment.getObjectMapper()),
+                    new IstioHeaderStorage()
+            )
         );
-        return new MetatronProxy(jsonClient, configuration.getMetatronUrl());
     }
 }
--- a/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
+++ b/proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
@@ -9,10 +9,10 @@ import io.dropwizard.configuration.SubstitutingSourceProvider;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import io.dropwizard.views.ViewBundle;
+import javax.ws.rs.client.Client;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
 import org.opensaml.core.config.InitializationException;
 import org.opensaml.core.config.InitializationService;
-
 import uk.gov.ida.common.shared.security.X509CertificateFactory;
 import uk.gov.ida.dropwizard.logstash.LogstashBundle;
 import uk.gov.ida.jerseyclient.ErrorHandlingClient;
@@ -38,13 +38,10 @@ import uk.gov.ida.notification.translator.saml.EidasFailureResponseGenerator;
 import uk.gov.ida.notification.translator.saml.EidasResponseGenerator;
 import uk.gov.ida.notification.translator.saml.HubResponseTranslator;
 
-import javax.ws.rs.client.Client;
-
 public class TranslatorApplication extends Application<TranslatorConfiguration> {
-
-    public static void main(final String[] args) throws Exception {
+    public static void main(String[] args) throws Exception {
         if (args == null || args.length == 0) {
-            String configFile = System.getenv("CONFIG_FILE");
+            final String configFile = System.getenv("CONFIG_FILE");
 
             if (configFile == null) {
                 throw new RuntimeException("CONFIG_FILE environment variable should be set with path to configuration file");
@@ -62,7 +59,7 @@ public class TranslatorApplication extends Application<TranslatorConfiguration> 
     }
 
     @Override
-    public void initialize(final Bootstrap<TranslatorConfiguration> bootstrap) {
+    public void initialize(Bootstrap<TranslatorConfiguration> bootstrap) {
         // Needed to correctly interpolate environment variables in config file
         bootstrap.setConfigurationSourceProvider(
                 new SubstitutingSourceProvider(bootstrap.getConfigurationSourceProvider(),
@@ -87,10 +84,10 @@ public class TranslatorApplication extends Application<TranslatorConfiguration> 
     }
 
     @Override
-    public void run(final TranslatorConfiguration configuration, final Environment environment) {
+    public void run(TranslatorConfiguration configuration, Environment environment) {
         environment.jersey().register(IstioHeaderMapperFilter.class);
         environment.jersey().register(ProxyNodeLoggingFilter.class);
-        ProxyNodeHealthCheck proxyNodeHealthCheck = new ProxyNodeHealthCheck("translator");
+        final ProxyNodeHealthCheck proxyNodeHealthCheck = new ProxyNodeHealthCheck("translator");
         environment.healthChecks().register(proxyNodeHealthCheck.getName(), proxyNodeHealthCheck);
 
         registerExceptionMappers(environment);
@@ -105,6 +102,15 @@ public class TranslatorApplication extends Application<TranslatorConfiguration> 
         environment.jersey().register(new CatchAllExceptionMapper());
     }
 
+    private void registerInjections(Environment environment) {
+        environment.jersey().register(new AbstractBinder() {
+            @Override
+            protected void configure() {
+                bindAsContract(IstioHeaderStorage.class);
+            }
+        });
+    }
+
     private void registerResources(TranslatorConfiguration configuration, Environment environment) {
         final EidasResponseGenerator eidasResponseGenerator = createEidasResponseGenerator(configuration, environment);
         final VerifyServiceProviderProxy vspProxy = configuration.getVspConfiguration().buildVerifyServiceProviderProxy(environment);
@@ -115,7 +121,7 @@ public class TranslatorApplication extends Application<TranslatorConfiguration> 
 
     private EidasResponseGenerator createEidasResponseGenerator(TranslatorConfiguration configuration, Environment environment) {
 
-        MetatronProxy metatronProxy = buildMetatronProxy(configuration, environment);
+        MetatronProxy metatronProxy = createMetatronProxy(configuration, environment);
 
         final HubResponseTranslator hubResponseTranslator = new HubResponseTranslator(
                 EidasResponseBuilder::instance,
@@ -129,8 +135,7 @@ public class TranslatorApplication extends Application<TranslatorConfiguration> 
 
         final CredentialConfiguration credentialConfiguration = configuration.getCredentialConfiguration();
         final SamlObjectSigner samlObjectSigner = new SamlObjectSigner(credentialConfiguration.getCredential(),
-                                                                       credentialConfiguration.getAlgorithm(),
-                                                                       credentialConfiguration.getKeyHandle());
+                credentialConfiguration.getAlgorithm(), credentialConfiguration.getKeyHandle());
 
         return new EidasResponseGenerator(
                 hubResponseTranslator,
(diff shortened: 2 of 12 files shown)


proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
```diff
@@ -29,25 +30,24 @@
 
     @Override
     public void filter(ContainerRequestContext requestContext) {
-        String istioTraceId = requestContext.getHeaderString(X_B3_TRACEID);
-        if (istioTraceId != null && !istioTraceId.isBlank()) {
-            MDC.put(X_B3_TRACEID, istioTraceId);
-        }
+        var istioTraceId = ofNullable(requestContext.getHeaderString(X_B3_TRACEID));
+        istioTraceId.filter(Predicate.not(String::isBlank)).ifPresent(traceId -> MDC.put(X_B3_TRACEID, traceId));
```
JohnWattsGDS: 😄

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
```diff
@@ -30,24 +26,26 @@
...
 import uk.gov.ida.notification.translator.resources.HubResponseTranslatorResource;
 import uk.gov.ida.notification.translator.saml.EidasFailureResponseGenerator;
 import uk.gov.ida.notification.translator.saml.EidasResponseGenerator;
 import uk.gov.ida.notification.translator.saml.HubResponseTranslator;
 
-import javax.ws.rs.client.Client;
-
 public class TranslatorApplication extends Application<TranslatorConfiguration> {
 
+    private MetatronProxy espMetatronProxy;
```
JohnWattsGDS: Are we able to use one MetatronProxy and pass a reference to the one proxy instead of creating two?

jkunle: I thought we introduced the Factory so we could have separate instances, one for the translator and one for the ESP

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
CharlesIC: What??

CharlesIC: Actually it's quite clever but I would just chain it instead of splitting into two statements

CharlesIC:
```
ofNullable(requestContext.getHeaderString(X_B3_TRACEID).filter(Predicate.not(String::isBlank)).ifPresent(traceId -> MDC.put(X_B3_TRACEID, traceId));
```

CharlesIC: Or it might even fit on one line

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
CharlesIC: I think it's fine to have two proxies for the different microservices - they're supposed to be separate and really we should be using DI but that's a story for another time

CharlesIC: Wait but why is `espMetatronProxy` in `TranslatorApplication`, though...? This doesn't seem right

CharlesIC: We certainly don't need two metatron proxies here - one is enough and totally fine. The translator only needs one instance - which will be used also by objects it creates.
Also the instance called `espMetatronProxy` here is NOT used by ESP - it's used by `EidasResponseGenerator` which is an object used by the translator. ESP is a separate microservice which is created by its own Application class - it's completely separate from the Translator.

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
```diff
@@ -30,24 +26,26 @@
...
+    private MetatronProxy translatorMetatronProxy;
```
CharlesIC:
```suggestion
    private MetatronProxy metatronProxy;
```
Just one is enough

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/MetatronProxyFactory.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.dropwizard.client.JerseyClientBuilder;
+import io.dropwizard.client.JerseyClientConfiguration;
+import io.dropwizard.setup.Environment;
+import uk.gov.ida.jerseyclient.ErrorHandlingClient;
+import uk.gov.ida.jerseyclient.JsonResponseProcessor;
+import uk.gov.ida.notification.shared.istio.IstioHeaderStorage;
+import javax.ws.rs.client.Client;
+
+public class MetatronProxyFactory {
```
CharlesIC: I don't think we need a factory to create a simple class like the Metatron proxy - it only takes two things to build, one of which is a URI. Put all the code in the `buildMetatronProxy` method into the Metatron proxy constructor instead and we can get rid of this class.

Actually we have proxies in a few places now and we should refactor this into a pattern (probably a base Proxy class and a bunch of children)

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserApplication.java
```diff
@@ -31,7 +31,7 @@
 import uk.gov.ida.notification.shared.proxy.MetatronProxy;
 import uk.gov.ida.notification.shared.proxy.MetatronProxyFactory;
 
-public class EidasSamlApplication extends Application<EidasSamlParserConfiguration> {
+public class EidasSamlParserApplication extends Application<EidasSamlParserConfiguration> {
```
CharlesIC: Yes to that! It's been bothering me for a long time, well spotted 😄

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/MetatronProxyFactory.java
JohnWattsGDS: The factory definitely turned out to be overkill but the only thing the proxies really have in common is a Client and a URL and they're just fields; there's no shared functionality or logic so having an underlying base class would also be overkill.

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
jkunle: You are right, i thought the variable name istioTraceId had some significance. thanks

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
jkunle: Thanks 👍

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/proxy/MetatronProxyFactory.java
jkunle: 👍

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
```diff
@@ -29,25 +30,24 @@
...
+        ofNullable(requestContext.getUriInfo()).ifPresent(u -> ProxyNodeLogger.addContext(RESOURCE_PATH, u.getAbsolutePath().toString()));
+        ofNullable(requestContext.getMediaType()).ifPresent(m -> ProxyNodeLogger.addContext(INGRESS_MEDIA_TYPE, m.toString()));
+
+        ProxyNodeLogger.info(MESSAGE_INGRESS);
         MDC.remove(INGRESS_MEDIA_TYPE.name());
     }
 
     @Override
     public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
-        Optional.ofNullable(responseContext.getLocation()).ifPresent(uri -> ProxyNodeLogger.addContext(EGRESS_LOCATION, uri.toString()));
```
idrop: I remember being told that we favoured the fully-qualified imports i.e. `Optional.ofNullable`. We should discuss this  at dev huddle.

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
```diff
@@ -29,25 +30,24 @@
 
     @Override
     public void filter(ContainerRequestContext requestContext) {
-        String istioTraceId = requestContext.getHeaderString(X_B3_TRACEID);
-        if (istioTraceId != null && !istioTraceId.isBlank()) {
-            MDC.put(X_B3_TRACEID, istioTraceId);
-        }
+        ofNullable(requestContext.getHeaderString(X_B3_TRACEID))
+            .filter(Predicate.not(String::isBlank)).ifPresent(traceId -> MDC.put(X_B3_TRACEID, traceId));
```
idrop: Complete nitpick, but the `Predicate.not` method is a fully-qualified import, whilst the rest of the line uses static imports. Can we make it a static import?

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -53,15 +55,22 @@ public Response hubResponse(@Valid final HubResponseTranslatorRequest hubRespons
...
+                                                        .map(AssertionConsumerService::getLocation)
+                                                        .map(URI::toString)
+                                                        .orElseThrow();
 
         EidasAuthnResponseAttributesHashLogger.logEidasAttributesHash(
                             translatedHubResponse.getAttributes().orElse(null),
                             translatedHubResponse.getPid().orElse(null),
                             hubResponseTranslatorRequest.getRequestId(),
-                            countryMetadata.getDestination());
+                            URI.create(assertionConsumerServiceURL));
```
CharlesIC: The assertion consumer service URL should be taken from the `hubResponseTranslatorRequest` (`destinationUrl`) - which includes the URL already preselected by the ESP. We mustn't repeat the logic here as we don't have access to the URL passed in from the `AuthnReuqest`

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -53,15 +55,22 @@ public Response hubResponse(@Valid final HubResponseTranslatorRequest hubRespons
             final String entityId = hubResponseTranslatorRequest.getEidasIssuerEntityId().toString();
             final TranslatedHubResponse translatedHubResponse = getAttributesFromVSP(hubResponseTranslatorRequest);
             final CountryMetadataResponse countryMetadata = metatronProxy.getCountryMetadata(entityId);
+            final String assertionConsumerServiceURL = countryMetadata.getAssertionConsumerServices()
```
CharlesIC: Don't duplicate logic - use the URL pre-selected by the ESP which is in the `hubResponseTranslatorRequest` as `destinationUrl`

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
```diff
@@ -52,7 +52,11 @@ public Response hubResponse(@Valid HubResponseTranslatorRequest hubResponseTrans
                 hubResponseTranslatorRequest.getRequestId(),
                 hubResponseTranslatorRequest.getDestinationUrl());
 
-        HubResponseContainer hubResponseContainer = new HubResponseContainer(hubResponseTranslatorRequest, translatedHubResponse);
+        HubResponseContainer hubResponseContainer = new HubResponseContainer(
+                hubResponseTranslatorRequest,
+                translatedHubResponse,
+                hubResponseTranslatorRequest.getDestinationUrl()
```
idrop: We are getting the destinationUrl here via the `ESP -> gateway -> redis -> gateway -> translator`.
I'm hoping we could just use the metatron to evaluate the destination url here instead, and simplify our DTOs even more. This evaluation would be exactly as done in `EidasSamlResource.post`

idrop: We want to respect the destinationUrl sent in the AuthnRequest, so some version of the data passing above still needs to happen. So please ignore this review.

eidas-saml-parser/src/main/java/uk/gov/ida/notification/eidassaml/EidasSamlParserApplication.java
```diff
@@ -128,14 +129,4 @@ protected void configure() {
             }
         });
     }
-
-    private MetatronProxy buildMetatronProxy(EidasSamlParserConfiguration configuration, Environment environment) {
```
CharlesIC: This method was okay to keep - it is useful but rename it to `createMetatronProxy ` (like we have for other things)

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
CharlesIC: Yes we do most of the time

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
```diff
@@ -29,25 +30,24 @@
 
     @Override
     public void filter(ContainerRequestContext requestContext) {
-        String istioTraceId = requestContext.getHeaderString(X_B3_TRACEID);
-        if (istioTraceId != null && !istioTraceId.isBlank()) {
-            MDC.put(X_B3_TRACEID, istioTraceId);
-        }
+        ofNullable(requestContext.getHeaderString(X_B3_TRACEID))
```
CharlesIC: Was there a reason to do this? This is much less clear so unless we have a good reason to use Optionals here this should be reverted. The `if` statement is more verbose but more clear and easier to understand.

proxy-node-shared/src/main/java/uk/gov/ida/notification/shared/logging/ProxyNodeLoggingFilter.java
```diff
@@ -29,25 +30,24 @@
...
+            .filter(not(String::isBlank)).ifPresent(traceId -> MDC.put(X_B3_TRACEID, traceId));
 
         ProxyNodeLogger.addContext(PROXY_NODE_JOURNEY_ID, getJourneyId(requestContext));
-        ProxyNodeLogger.addContext(REFERER, Optional.ofNullable(requestContext.getHeaderString(HttpHeaders.REFERER)).orElse(""));
-        Optional.ofNullable(requestContext.getUriInfo()).ifPresent(u -> ProxyNodeLogger.addContext(RESOURCE_PATH, u.getAbsolutePath().toString()));
```
CharlesIC: We should also keep the `Optional` in `Optional.ofNullable` as in this case it's less clear without the full name

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
```diff
@@ -107,15 +103,23 @@ private void registerExceptionMappers(Environment environment) {
 
     private void registerResources(TranslatorConfiguration configuration, Environment environment) {
         final EidasResponseGenerator eidasResponseGenerator = createEidasResponseGenerator(configuration, environment);
-        final VerifyServiceProviderProxy vspProxy = configuration.getVspConfiguration().buildVerifyServiceProviderProxy(environment);
-        final HubResponseTranslatorResource hubResponseTranslatorResource = new HubResponseTranslatorResource(eidasResponseGenerator, vspProxy);
+        final VerifyServiceProviderProxy vspProxy = configuration.getVspConfiguration()
```
CharlesIC: This can be on one line

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
```diff
@@ -107,15 +103,23 @@ private void registerExceptionMappers(Environment environment) {
...
+                .buildVerifyServiceProviderProxy(environment);
+        final HubResponseTranslatorResource hubResponseTranslatorResource = new HubResponseTranslatorResource(
```
CharlesIC: This can be on one line

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/TranslatorApplication.java
```diff
@@ -149,15 +147,4 @@ protected void configure() {
             }
         });
     }
-
-    public MetatronProxy buildMetatronProxy(TranslatorConfiguration configuration, Environment environment) {
```
CharlesIC: We can keep this method as it's is useful, just rename it to `createMetatronProxy` like we have for other things

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/resources/HubResponseTranslatorResource.java
CharlesIC: Yep getting this from `hubResponseTranslatorRequest` is correct as it has been selected by the ESP taking into account also the authn request - it would be nice to refactor this but for now this is the right thing to do

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/EidasResponseGenerator.java
```diff
@@ -22,12 +22,9 @@
     private final MetatronProxy metatronProxy;
     private final X509CertificateFactory x509CertificateFactory;
 
-    public EidasResponseGenerator(
-            HubResponseTranslator hubResponseTranslator,
-            EidasFailureResponseGenerator failureResponseGenerator,
-            SamlObjectSigner samlObjectSigner,
-            MetatronProxy metatronProxy,
-            X509CertificateFactory x509CertificateFactory) {
+    public EidasResponseGenerator(final HubResponseTranslator hubResponseTranslator,
```
CharlesIC: All of these should be on separate lines - this seems to be the convention we use when constructors have many arguments. I can see a few different styles in this PR - we should be consistent with style as much as possible

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/EidasResponseGenerator.java
```diff
@@ -36,33 +33,28 @@ public EidasResponseGenerator(
     }
 
     public Response generateFromHubResponse(final HubResponseContainer hubResponseContainer) {
-        final CountryMetadataResponse countryMetadataResponse = metatronProxy.getCountryMetadata(hubResponseContainer.getIssuer().toString());
-        final Response eidasResponse = hubResponseTranslator.getTranslatedHubResponse(hubResponseContainer, countryMetadataResponse);
-        final Response encryptedEidasResponse = encryptAssertions(eidasResponse, countryMetadataResponse.getSamlEncryptionCertX509());
+        final CountryMetadataResponse countryMetadataResponse = metatronProxy
```
CharlesIC: this can be on one line

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/EidasResponseGenerator.java
```diff
@@ -36,33 +33,28 @@ public EidasResponseGenerator(
...
+                .getCountryMetadata(hubResponseContainer.getIssuer().toString());
+        final Response eidasResponse = hubResponseTranslator.getTranslatedHubResponse(hubResponseContainer,
```
CharlesIC: this can be on one line

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/EidasResponseGenerator.java
```diff
@@ -36,33 +33,28 @@ public EidasResponseGenerator(
...
+                countryMetadataResponse);
+        final Response encryptedEidasResponse = encryptAssertions(eidasResponse,
```
CharlesIC: this can be on one line

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/EidasResponseGenerator.java
```diff
@@ -36,33 +33,28 @@ public EidasResponseGenerator(
...
-        Response eidasResponse = failureResponseGenerator.generateFailureSamlResponse(
-                responseStatus,
-                eidasRequestId,
-                destinationUrl,
-                entityId.toString()
-        );
+    public Response generateFailureResponse(final javax.ws.rs.core.Response.Status responseStatus,
+            final String eidasRequestId, final String destinationUrl, final URI entityId) {
 
+        final Response eidasResponse = failureResponseGenerator.generateFailureSamlResponse(responseStatus,
```
CharlesIC: move `responseStatus` to the next line with the rest of the args

proxy-node-translator/src/main/java/uk/gov/ida/notification/translator/saml/HubResponseContainer.java
```diff
@@ -11,18 +11,21 @@
...
+    private final VspScenario vspScenario;
+    private final VspLevelOfAssurance levelOfAssurance;
+
+    public HubResponseContainer(
+            final HubResponseTranslatorRequest hubResponseTranslatorRequest,
+            final TranslatedHubResponse translatedHubResponse,
+            final URI destination) {
         this.pid = translatedHubResponse.getPid().orElse(null);
         this.eidasRequestId = hubResponseTranslatorRequest.getEidasRequestId();
-        this.destinationURL = hubResponseTranslatorRequest.getDestinationUrl();
```
CharlesIC: `hubResponseTranslatorRequest.getDestinationUrl()` is fine - the `destinationUrl` is contained in `hubResponseTranslatorRequest` and should not be a separate argument of the constructor - this is confusing

PR #512, opened 2020-02-25
EID-1917: Add smoke tests to all countries connected to Connector Node

AmritSidhu:
The tests all start from the 'Personal tax account' service. From there sign in with a digital identity from another European country is selected. Then the country of choice is selected and then we check that we have reached the countries hub by asserting we can see specialised text on the page.

The smoke test will be run by the hub smoke test pipeline.

These feature can be tested locally by:

- updating the ENTRYPOINT in proxy-node-acceptance-tests/Dockerfile to:
`ENTRYPOINT ["./wait-for-it.sh", "http://selenium-hub:4444/status", "--", "bundle", "exec", "cucumber", "features/smoke/{insert the country abbreviation}-request.feature", "--strict", "--tags", "not @ignore"]`

- running ./proxy-node-acceptance-tests/run-staging-tests.sh

Add a smoke test for each country to the Verify Hub Smoke pipeline: https://github.com/alphagov/verify-terraform/pull/1162

### DIFF
--- a/proxy-node-acceptance-tests/features/smoke/es-request.feature
+++ b/proxy-node-acceptance-tests/features/smoke/es-request.feature
@@ -1,0 +1,7 @@
+Feature: eidas-proxy-node-smoke-test-es-prod feature
+
+    Scenario: Connector node happy path for Spain
+        Given   the user visits a government service
+        And     they choose sign in with a digital identity from another European country
+        And     they select Spain
+        Then    they should arrive at the Spain Hub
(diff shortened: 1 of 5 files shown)


Wynndow:
I'm not sure this is the correct place for these tests. These tests are checking our connection with other countries via our connector node, not our proxy node.

We currently have smoke tests for the middleware, are they're defined here: https://github.com/alphagov/ida-hub-acceptance-tests/blob/master/src/test/java/uk/gov/ida/acceptance/smoke/SmokeTests.java#L115

Having said that, I'm not convinces they should go in there either. Currently if the middleware goes down our smoke tests fails, which means we can't release the hub. This doesn't make a lot of sense. Adding more dependency on other countries being up and running for us to release the hub seems like a bad idea...

PR #514, opened 2020-02-25
EID-1926: create a sealed secret for non-prod vsp

jkunle:
Co-authored-by: Phillip Miller <phillip.miller@digital.cabinet-office.gov.uk>


ci/build/vsp-secret.yaml
```diff
@@ -1,12 +1,12 @@
...
+    type: Opaque
```
idrop: Think this indentation should be corrected

ci/integration/vsp-secret.yaml
```diff
@@ -1,12 +1,13 @@
...
+    type: Opaque
```
idrop: Think this indentation should be corrected, as above

PR #522, opened 2020-02-28
[DO NOT CLOSE] Catch-all PR for Sandbox to verify signatures

CharlesIC:
This PR is required due to what is arguably a bug in concourse-github-resource.

This PR triggers GitHub to verify signatures so the custom GSP resource can successfully verify commits. This hack is needed to enable us to push to Sandbox directly without having to open PRs. We need to do this until the bug in the resource is fixed.


Krenair:
PR should no longer be required given https://github.com/alphagov/gsp/pull/1053 + https://github.com/alphagov/tech-ops-cluster-config/pull/297

PR #531, opened 2020-04-08
✎ Document how Proxy Node Metadata signing works

idrop:
Before starting work on creating a new chain of trust for multi-country Proxy Node, we should have some idea of how the chain of trust works.


doc/metadata_signing.md
```diff
@@ -0,0 +1,43 @@
...
+ [Proxy Node metadata](https://proxy-node.eidas.test.london.verify.govsvc.uk/ServiceMetadata) is generated by the [Verify Metadata Controller](https://github.com/alphagov/verify-metadata-controller), from a kubeyaml declaration of
+`kind: Metadata`.
+
+The metadata is signed with a CloudHSM key and its accompanying metadata signing certificate is chained to the following roots of trust in prodution or test:
+
+````
+Production
+├── verify-root-ca-a1
+│   └── proxy-node-metadata-ca-a1
+│       └── verify-eidas-proxy-node-prod-metadata-signing-cert-a1
```
CharlesIC, line 12: Is it always going to be a1?

doc/metadata_signing.md
```diff
@@ -0,0 +1,43 @@
...
+ ${namespace}-${secret}:public
+````
+for the private and public key respectively. These CloudHSM keys cannot be deleted.
+
+## Deleting and generating certificates
+Certificates can be deleted by deleting the Kubernetes secret that holds them.
+
+If the kubeyaml of `kind: CertificateRequest` is retained in the cluster namespace with the same name, the certificate and secret will be regenerated using the existing CloudHSM keypair.
+
+The `verify-metadata-controller` pod `v1-vmc-0` can be cycled for this regeneration.
```
CharlesIC, line 40: Is it always going to be `v1-vmc-0` (pod 0 specifically)?

doc/metadata_signing.md
idrop: It will still be `a1` until we want to use a new EC keypair. The keypair for `verify-root-ca-a1` for instance can't be deleted with our current codebase.
The certificates signed by these keys can be regenerated when expiring or necessary. If we want a new keypair for reasons, then we'd have to switch the tree to `a2` from the root down.

doc/metadata_signing.md
idrop: It's always been that name since it existed so seems safe to refer to it by name. I can make it more general by referring to the vmc pod though.

doc/metadata_signing.md
```diff
@@ -0,0 +1,43 @@
+#Proxy Node Metadata Signing
+
+ [Proxy Node metadata](https://proxy-node.eidas.test.london.verify.govsvc.uk/ServiceMetadata) is generated by the [Verify Metadata Controller](https://github.com/alphagov/verify-metadata-controller), from a kubeyaml declaration of
+`kind: Metadata`.
+
+The metadata is signed with a CloudHSM key and its accompanying metadata signing certificate is chained to the following roots of trust in prodution or test:
```
jkunle, line 6: typo in **production**

idrop: Cheers, fixed 👍

AmritSidhu (approved):

PR #539, opened 2020-04-16
EID-1894: Add additional yaml files

AmritSidhu:
Copied artefact-signing-key-secret.yaml from ci/prod /integration. The artefact signing key is used to gpg sign the release that gets pushed to github

As we are only deploying (and not building) you don't need any of the extra files such as the principal.yaml and the image-repositories.yaml


ci/deploy/image-repositories.yaml
```diff
@@ -0,0 +1,90 @@
+---
```
blairboy362, line 1: If you're only deploying (and not building) you don't need any of these (see ci/prod or ci/integration)

ci/deploy/principal.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
blairboy362, line 1: If you're only deploying (and not building) you don't need this (see ci/prod or ci/integration)

PR #546, opened 2020-04-21
🚭BAU remove smoke tests from build acceptance tests

idrop:
We currently run all cucumber feature tests in our build pipeline, including tests in features/smoke that should be run only from the Hub smoke pipeline.

In the build pipeline, run only the features/proxy_node.feature, which covers build acceptance tests.

Tested locally with:
`./run-staging-tests.sh`


CharlesIC:
This would work but can we instead exclude the `smoke` directory?

CharlesIC:
It would be better to tag the smoke tests with `@smoke` and exclude that tag

PR #550, opened 2020-04-24
EID-1894 Include namespace in Proxy Node URLs

idrop:
Our deploy namespace `verify-eidas-proxy-node-deploy` needs to be part of our DNS name, which needs to end with `.<namespace>.`london.verify.govsvc.uk.

We have this constraint now so people who are allowed to administer other specific namespaces can't usurp our production system.

Add `.Release.Namespace` to gateway and connector hostnames to accommodate this.

This will affect the `verify-eidas-proxy-node-build` namespace urls too, expect further changes to docs and egress safelists.

We'll create CNAME records under `signin.service.gov.uk` to these urls.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -857,8 +857,8 @@ spec:
         config:
           platform: linux
           params:
-            PROXY_NODE_URL: "https://proxy-node.eidas.test.((cluster.domain))"
-            STUB_CONNECTOR_URL: "https://stub-connector.eidas.test.((cluster.domain))"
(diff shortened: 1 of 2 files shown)


ci/build/build-pipeline.yaml
```diff
@@ -857,8 +857,8 @@ spec:
         config:
           platform: linux
           params:
-            PROXY_NODE_URL: "https://proxy-node.eidas.test.((cluster.domain))"
-            STUB_CONNECTOR_URL: "https://stub-connector.eidas.test.((cluster.domain))"
+            PROXY_NODE_URL: "https://proxy-node.test.((namespace-deployer.namespace)).((cluster.domain))"
```
CharlesIC, line 860: Nice! 👍

PR #556, opened 2020-05-11
EID-1964 Use gov.uk domain for external urls

idrop:
Use the domain `eidas.signin.service.gov.uk` for external urls and entity ids.

FYI: We'll have to do something special in `_helpers.tpl` for `sandbox` which won't have gov.uk urls.


CharlesIC:
yep we'll need to raise a ticket with RE to implement a custom domain for Sandbox as well

Krenair:
Uh this looks potentially problematic, you still need to respond to govsvc.uk requests or our proxy from eidas.signin.service.gov.uk won't work

chrisfarms:
Yeah the way @Krenair has the cloudfront setup, your application still has to "work" via HTTPS on the cluster domain.

Changing your gateway.host template will change the virtualservice/gateway/certificates ... which will stop everything working: 

* https://github.com/alphagov/verify-proxy-node/blob/a41732024b44f535089a75f773e72e776d196006/chart/templates/gateway-ingress-virtual-service.yaml#L13
* https://github.com/alphagov/verify-proxy-node/blob/master/chart/templates/gateway-ingress-certificate.yaml#L15

I think the only parts you need to change in your application is things like redirect urls etc, anything that shows/returns the URL to the public

PR #557, opened 2020-05-12
BAU Snyk vulnerability in hibernate-validator

idrop:
Address a vulnerability indicated in https://snyk.io/vuln/SNYK-JAVA-ORGHIBERNATE-568162

Cannot be applied to `configurations.all` as it results in NoClassDefFoundErrors.

The better solution is to upgrade to Dropwizard 2 which omits the vuln, but this is not a smooth migration.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,6 +108,14 @@ subprojects {
             }
         }
 
+        ['ida_utils','verify_saml','dropwizard','dropwizard_assets'].each {
+            configurations.getByName(it) {
+                resolutionStrategy.dependencySubstitution {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -108,6 +108,14 @@ subprojects {
             }
         }
 
+        ['ida_utils','verify_saml','dropwizard','dropwizard_assets'].each {
+            configurations.getByName(it) {
```
CharlesIC, line 112: What's the `it` referring to?

JohnWattsGDS: I think it's the each of the eaches. 😆 Or an implicit parameter.

https://groovy-lang.org/closures.html

One day I'm going to do a project and use Groovy for the testing because there'a s really cool library the name of which I've forgotten but it looked great and used all of those groovy features.

JohnWattsGDS: This is the library: https://www.baeldung.com/groovy-spock

PR #558, opened 2020-05-12
EID-1894: add smoketest for multi-country integration proxy node

AmritSidhu:
I created a production and integration directory
Moved the Netherlands smoketest into the production directory. Subsequent PR in verify-terraform to also reflect the directory change - https://github.com/alphagov/verify-terraform/pull/1203
Copied over the existing integration test and put it into the integration directory as it does what we need it to do.

This needs to be merged in before we can play around with fly to test to see if it actually works


proxy-node-acceptance-tests/features/smoke/integration/integration.feature
```diff
@@ -0,0 +1,7 @@
+Feature: eidas-proxy-node-smoke-test-integration feature
+
+    Scenario: Integration proxy node happy path - LOA Substantial
+        Given the proxy node is sent a LOA 'Substantial' request
```
galund, line 4: Don't you need something here that suggests a stub, or anything that forces this to be integration rather than prod?

AmritSidhu: So in the steps the 'Given the proxy node is sent a LOA 'Substantial'...' will `visit(ENV.fetch('STUB_CONNECTOR_URL') + loa_url)` and that STUB_CONNECTOR_URL is passed in the verify-terraform

galund: Ah, ok if that's _always_ a stub then perhaps it could be '..request from the stub connector' because then it's more clear?

PR #572, opened 2020-06-03
EID-1949 Accept eIDAS compliant signing algos in ESP

idrop:
Support the following AuthnRequest signing algos within the ESP:

http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1
http://www.w3.org/2007/05/xmldsig-more#sha384-rsa-MGF1
http://www.w3.org/2007/05/xmldsig-more#sha512-rsa-MGF1
http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256
http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha384
http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha512

by using BouncyCastleProvider as a SecurityProvider in the ESP.

Add an integration test to check that signing algo http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1 is supported.

Make http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1 the default signing algo used in stub-connector.

### DIFF
--- a/proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
+++ b/proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
@@ -6,10 +6,16 @@ import org.opensaml.saml.common.SAMLVersion;
 import org.opensaml.saml.saml2.core.AuthnContextComparisonTypeEnumeration;
 import org.opensaml.saml.saml2.core.AuthnRequest;
 import org.opensaml.security.credential.Credential;
+import org.opensaml.xmlsec.algorithm.SignatureAlgorithm;
 import org.opensaml.xmlsec.algorithm.descriptors.SignatureRSASHA256;
 import org.opensaml.xmlsec.signature.support.SignatureException;
 import org.opensaml.xmlsec.signature.support.Signer;
-import org.w3c.dom.*;
+import org.w3c.dom.Attr;
+import org.w3c.dom.DOMException;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
 import uk.gov.ida.notification.saml.SamlParser;
 import uk.gov.ida.saml.core.test.builders.SignatureBuilder;
 
@@ -18,6 +24,7 @@ import javax.xml.xpath.XPathExpressionException;
 import java.text.MessageFormat;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 
 public class EidasAuthnRequestBuilder {
@@ -30,6 +37,7 @@ public class EidasAuthnRequestBuilder {
     private final String eidas = "http://eidas.europa.eu/saml-extensions";
     private HashMap<String, String> namespaceMap;
     private Credential signingCredential;
+    private SignatureAlgorithm signatureAlgorithm;
 
     public EidasAuthnRequestBuilder() throws Exception {
         parser = new SamlParser();
@@ -48,7 +56,7 @@ public class EidasAuthnRequestBuilder {
         if (this.signingCredential != null) {
             SignatureBuilder signatureBuilder = SignatureBuilder
                     .aSignature()
-                    .withSignatureAlgorithm(new SignatureRSASHA256())
+                    .withSignatureAlgorithm(Optional.ofNullable(signatureAlgorithm).orElseGet(SignatureRSASHA256::new))
                     .withSigningCredential(this.signingCredential);
             authnRequest.setSignature(signatureBuilder.build());
             XMLObjectProviderRegistrySupport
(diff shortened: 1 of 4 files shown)


proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
```diff
@@ -6,6 +6,7 @@
 import org.opensaml.saml.saml2.core.AuthnContextComparisonTypeEnumeration;
 import org.opensaml.saml.saml2.core.AuthnRequest;
 import org.opensaml.security.credential.Credential;
+import org.opensaml.xmlsec.algorithm.SignatureAlgorithm;
 import org.opensaml.xmlsec.algorithm.descriptors.SignatureRSASHA256;
 import org.opensaml.xmlsec.signature.support.SignatureException;
 import org.opensaml.xmlsec.signature.support.Signer;
```
jkunle, line 12: import org.w3c.dom.*;

proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
```diff
@@ -48,7 +51,7 @@ public AuthnRequest build() throws TransformerException, SignatureException, Mar
         if (this.signingCredential != null) {
             SignatureBuilder signatureBuilder = SignatureBuilder
                     .aSignature()
-                    .withSignatureAlgorithm(new SignatureRSASHA256())
+                    .withSignatureAlgorithm(Optional.ofNullable(signatureAlgorithm).orElse(new SignatureRSASHA256()))
```
jkunle, line 54: I think the logic should be inside the definition of `withSignatureAlgorithm`
   ``` 
public EidasAuthnRequestBuilder withSignatureAlgorithm(SignatureAlgorithm signatureAlgorithm) {
        this.signatureAlgorithm = signatureAlgorithm == null ? new SignatureRSASHA256() :  signatureAlgorithm ;
        return this;
 }
```
I don't think we need optional here as well

Comment (instantiator) [inline proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java:54]: Looks good - and keeps the default behaviour. Should it? Or at some point do we want this to fail if the signature algorithm isn't explicitly specified?

File: stub-connector/src/main/java/uk/gov/ida/notification/stubconnector/ConnectorNodeCredentialConfiguration.java [hunk 3]
```diff
@@ -44,7 +45,7 @@ public ConnectorNodeCredentialConfiguration(
         if (samlSigningCredential.getPublicKey() instanceof ECPublicKey) {
             this.algorithm = SignatureConstants.ALGO_ID_SIGNATURE_ECDSA_SHA256;
         } else {
-            this.algorithm = SignatureConstants.ALGO_ID_SIGNATURE_RSA_SHA256;
+            this.algorithm = XMLSignature.ALGO_ID_SIGNATURE_RSA_SHA256_MGF1;
```
instantiator, line 48: This is interesting - and I assume it's simply because `RSA_SHA256` isn't compliant but `RSA_SHA256_MGF1` is?

proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
jkunle: I understand why you've done this, I guess it is fine.

jkunle: May be this instead
` ``               .withSignatureAlgorithm(Optional.ofNullable(signatureAlgorithm).orElseGet(SignatureRSASHA256::new))
```

idrop: Pushed a change for this

proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
idrop: Think the GDS dev guide indicates not to do wildcard import?

proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
idrop: We don't have any logic in the ESP to fail validation if the signing algo is outside the intersection of key length & required algo. 
Was going to talk to @CharlesIC about this

proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java
jkunle: There is a wild card import on L13 https://github.com/alphagov/verify-proxy-node/blob/f5960ff016b26443f8330c3fce4df024c51202fb/proxy-node-test/src/main/java/uk/gov/ida/notification/helpers/EidasAuthnRequestBuilder.java#L13
couldn't highlight it with the + key

idrop: Thanks, fixed 👍

PR #575, opened 2020-06-05
🎢 Separate deploy and test jobs in build pipeline

idrop:
Separate the deploy of the proxy node build from the acceptance tests which follow.

This has two benefits:
1) re-run only acceptance tests through concourse if they fail in a triggered build.
2) prepare for ZDD - another new zdd test job should run in parallel to the deploy job, and acceptance tests should run after deploy.

Tested the pipeline locally 👇 showing separate `deploy` and `test` jobs :
Screenshot 2020-06-05 at 16 17 37

It currently looks like:
Screenshot 2020-06-05 at 16 25 01

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -197,13 +197,13 @@ spec:
 
     groups:
       - name: deployment
-        jobs: [build-vsp-image, build-proxy-node, package, test, release, delete]
+        jobs: [build-vsp-image, build-proxy-node, package, deploy, test, release, delete]
       - name: proxy-node
-        jobs: [build-proxy-node, package, test, release, delete]
+        jobs: [build-proxy-node, package, deploy, test, release, delete]
       - name: components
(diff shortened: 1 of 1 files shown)


CharlesIC (commented):
Why remove the dependency `passed build-proxy-node` for the test image?
(I know it looks better and it's probably fine but want to check if there's any specific reason)

idrop:
> Why remove the dependency `passed build-proxy-node` for the test image?
> (I know it looks better and it's probably fine but want to check if there's any specific reason)

I chose to enforce build order on the `chart` resource (using the `passed` flag)  as its used in each job except `test` where its passed through. 
The way I see it, we don't need to carry the test-image through `delete` and `deploy` jobs.

chrisfarms (dismissed):

idrop:
I've had to rebase, re-review please 🙏

JohnWattsGDS (approved):

jkunle (approved):
LGTM

PR #580, opened 2020-06-18
EID-1951: Add integration smoketests for Estonia and Sweden

CharlesIC:
Improve Cucumber tests
  - Refactor country specific bits into separate methods for reusability
  - Improve spacing and capitalisation to make tests looks clean and tidy

Add smoke tests for Sweden and Estonia
  - Move country specific methods to a separate file for better organisation
  - Call country methods by constructing their names

### DIFF
--- a/proxy-node-acceptance-tests/features/step_definitions/country_journeys.rb
+++ b/proxy-node-acceptance-tests/features/step_definitions/country_journeys.rb
@@ -1,0 +1,60 @@
+# Start pages
+
+def country_stub_connector_url(country)
+  case country
+  when 'Netherlands'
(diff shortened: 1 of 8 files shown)


proxy-node-acceptance-tests/features/step_definitions/country_journeys.rb
```diff
@@ -0,0 +1,60 @@
+# Start pages
+
+def country_stub_connector_url(country)
+  case country
```
idrop, line 4: How about a map instead of a switch block?

CharlesIC: Why? The point of a switch is exactly this use case. We don't need to retain the map or anything

PR #581, opened 2020-06-22
Make smoke tests naming consistent with Connector Node smoke tests

CharlesIC:
To make them consistent with Connector Node tests and enable country status reporting the Proxy Node from alphagov/verify-terraform#1231

  - Rename to use country names

See also alphagov/verify-acceptance-tests#93

### DIFF
--- a/proxy-node-acceptance-tests/features/smoke/production/netherlands.feature
+++ b/proxy-node-acceptance-tests/features/smoke/production/netherlands.feature
@@ -1,0 +1,6 @@
+Feature: eIDAS Proxy Node Smoke Test - Netherlands - Production
+
+    Scenario: Proxy Node Netherlands happy path - LOA Substantial
+        Given   the user visits the "Netherlands" Stub Connector Node page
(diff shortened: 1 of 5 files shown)


proxy-node-acceptance-tests/features/smoke/production/netherlands.feature
```diff
@@ -0,0 +1,6 @@
+Feature: eIDAS Proxy Node Smoke Test - Netherlands - Production
+
+    Scenario: Proxy Node Netherlands happy path - LOA Substantial
+        Given   the user visits the "Netherlands" Stub Connector Node page
+        And     they navigate the "Netherlands" journey to verify with UK identity
+        Then    they should arrive at the Verify Hub start page
```
jkunle, line 6: No newline at the end of file could this be an issue?

CharlesIC: I've run the tests locally, Cucumber doesn't complain and the tests pass, so I think it's fine

PR #582, opened 2020-06-23
timestamp pods during helm templating

idrop:
## What
Ensure pod kubeyaml changes each deployment by changing the pod release timestamp.
Pods will then be deleted and rescheduled at deploy time by the deployment controller.

Also remove the `talksToHSM` label from stub-connector, because it doesn't anymore.

## Why
We are currently changing the timestamp on the deployment only, which means pods never 'roll'.
We rely on pods being rolled regularly to pick up the latest Metadata and SAML signing certs updated by the [VMC](https://github.com/alphagov/verify-metadata-controller).
We also release growing HSM key handles held by HSM client.

We have a [daily job](https://ci.london.verify.govsvc.uk/teams/eidas-proxy-node-deploy/pipelines/deploy) which will apply these changes.

See zendesk https://govuk.zendesk.com/agent/tickets/4134808 for symptoms.

### DIFF
--- a/chart/templates/esp-deployment.yaml
+++ b/chart/templates/esp-deployment.yaml
@@ -8,8 +8,6 @@ metadata:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
-  annotations:
-    timestamp: {{ .Release.Time }}
 spec:
   replicas: {{ .Values.esp.replicaCount }}
   strategy:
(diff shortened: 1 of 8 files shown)


idrop:
> Totes believe you that this will work. Would be nice if we could reference some docs or something in the commit message, so that people can learn why it's done this way from looking at the commit history.

@galund Good point, I've referenced the zendesk ticket that kicked this ticket off.

galund:
Thanks; the PR isn't a permanent thing in the history in the same way as the commit message, so do prefer the commit messages to be comprehensive if possible. (Also docs are better than tickets, tho tickets are better than nothing :-) )

PR #591, opened 2020-07-15
BAU fix proxy node build acceptance tests

idrop:
Build pipeline acc tests require a  TEST_ENV env var, and will fallback to local if not present.

Pass in the TEST_ENV env var of build as a key to environment.yaml.

### DIFF
--- a/ci/build/build-pipeline.yaml
+++ b/ci/build/build-pipeline.yaml
@@ -859,10 +859,7 @@ spec:
         config:
           platform: linux
           params:
-            PROXY_NODE_URL: "https://proxy-node.test.eidas.signin.service.gov.uk"
-            STUB_CONNECTOR_URL: "https://stub-connector.test.eidas.signin.service.gov.uk"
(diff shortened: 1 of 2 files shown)


ci/build/build-pipeline.yaml
```diff
@@ -859,10 +859,7 @@ spec:
         config:
           platform: linux
           params:
-            PROXY_NODE_URL: "https://proxy-node.test.eidas.signin.service.gov.uk"
-            STUB_CONNECTOR_URL: "https://stub-connector.test.eidas.signin.service.gov.uk"
-            STUB_IDP_USER: "stub-idp-demo-one"
```
JohnWattsGDS, line 864: What's happened to `STUB_IDP_USER`? Was it redundant?

idrop: The ruby script sets it to `stub-idp-demo-one` if not provided, so thought it wasn't necessary here

JohnWattsGDS: 👍

PR #593, opened 2020-07-16
BAU Fix Acceptance tests

idrop:
Use the `env` function to assign a var first from `environments.yml`, then from `ENV`.

This will pick up the `SELENIUM_HUB_URL` entry in `environments.yml` and allow the build pipeline to use the Selenium grid.

This PR is a consequence of https://github.com/alphagov/verify-proxy-node/pull/591 which fixed usage of `TEST_ENV` but broke usage of `SELENIUM_HUB_URL`.

Tested against integration, build and local proxy node. You can emulate the pipeline behaviour by doing:
`cd proxy-node-acceptance-tests`
`docker build . -t acctest`
`docker run -e TEST_ENV=build acctest`


JohnWattsGDS (approved):

proxy-node-acceptance-tests/features/support/env.rb
```diff
@@ -2,13 +2,21 @@
 require 'capybara-screenshot/cucumber'
 require 'selenium/webdriver'
 require 'webdrivers/chromedriver'
+require 'yaml'
+
+ENVIRONMENTS = YAML.load_file(File.join(__dir__, 'environments.yml'))
+
+def env(key)
+  test_env = ENV.fetch('TEST_ENV', 'local')
```
CharlesIC, line 10: I would initialise test_env separately outside this method and use it here as it's going to be fetched multiple times otherwise. Not a huge performance consideration but a more "cleaner" approach. Just like we initialise a bunch of vars below.

CharlesIC (approved):

proxy-node-acceptance-tests/features/support/env.rb
JohnWattsGDS: You can achieve the same effect with `test_env ||= ENV.fetch('TEST_ENV', 'local')` - welcome to the wonderful world of Ruby, Phil.

PR #602, opened 2020-07-21
EID-1960 Remember to delete the config value

JohnWattsGDS:
Delete proxyNodeMetadataForConnectorNodeUrl from the Configuration class and from the yaml file too.

As this is no longer an env var for k8s the value is never interpolated and Java tries to parse ${PROXY_NODE_METADATA_FOR_CONNECTOR_NODE_URL} as a URL.  Unsurprisingly this doesn't end well and the translator doesn't start at all which means tests never get run.

### DIFF
--- a/proxy-node-translator/src/dist/config.yml
+++ b/proxy-node-translator/src/dist/config.yml
@@ -16,8 +16,6 @@ logging:
   appenders:
     - type: ${LOGGING_APPENDER:-logstash-console}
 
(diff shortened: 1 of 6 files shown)


CharlesIC:
does this break local running?

JohnWattsGDS:
I don't think it will break any local running as it just removes values from configuration.  There won't even be any loose env vars hanging around as the connector uses the same value.

PR #603, opened 2020-07-22
EID-1960 Concourse is unhappy

JohnWattsGDS:
Concourse is complaining about signatures.  This might be because I re-opened a PR?  It seems bonkers but I'm clutching at straws.  So this PR is a small whitespace change that will hopefully convince Concourse that it's okay to go ahead and build what's in master.


JohnWattsGDS:
Concourse is happy again for no apparent reason so closing and deleting branch.

PR #604, opened 2020-07-23
EID-1919 Add sealed secret containing QWAC Cert

JohnWattsGDS:
Add some kubeyaml that should make the sealed secret available for the gateway to use in a later PR.


idrop:
Could you prefix the filename with `production` please, as it's only for prod and not integration?

idrop:
Sorry, one more thing: the name `name: gateway-ingress-certificate-for-signin-sealed-secret` could possibly be shortened to exclude `-sealed-secret`, by convention.

PR #605, opened 2020-07-27
EID-1919 Configure Gateway to use QWAC

JohnWattsGDS:
add `qwac.enabled: false` to `values.yaml` so that the default is to not use QWAC

add conditional sections to `gateway-ingress-gateway.yaml` under the `tls` and `hosts` sections so that when `qwac.enabled` is true, `tls` is handled differently and a different hostname is used.

add `--set "gateway.qwac=true"` to inline bash used in `render-manifests` in `prod/depoly-pipeline.yaml` so that production uses the QWAC

cross fingers and hope for the best.

### DIFF
--- a/chart/templates/gateway-ingress-gateway.yaml
+++ b/chart/templates/gateway-ingress-gateway.yaml
@@ -26,6 +26,15 @@ spec:
       number: 443
       name: https
       protocol: HTTPS
+{{- if .Values.global.qwacCertificate.enabled -}}
+    tls:
+      mode: SIMPLE
+      serverCertificate: sds
+      privateKey: sds
+      credentialName: gateway-ingress-certificate-for-signin-sealed-secret
+    hosts:
+    - {{ include "gateway.host.govuk" . }}
+{{- else -}}
     tls:
       mode: SIMPLE
       serverCertificate: sds
@@ -33,3 +42,4 @@ spec:
       credentialName: {{ .Release.Name }}-gateway-ingress-certificate
     hosts:
     - {{ include "gateway.host" . }}
+{{- end -}}
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -6,6 +6,8 @@ global:
     enabled: true
     ip: 127.0.0.1
   domainGovUK: eidas.signin.service.gov.uk
+  qwacCertificate:
+    enabled: false
 
 hubFqdn: www.integration.signin.service.gov.uk
 
--- a/ci/deploy/deploy-pipeline.yaml
+++ b/ci/deploy/deploy-pipeline.yaml
@@ -303,6 +303,7 @@ spec:
                     --set "global.cluster.name=${CLUSTER_NAME}" \
                     --set "global.cluster.domain=${CLUSTER_DOMAIN}" \
                     --set "global.cloudHsm.ip=${CLOUDHSM_IP}" \
+                    --set "global.qwacCertificate.enabled=true" \
                     --set "stubConnector.enabled=false" \
                     --set "vsp.secretName=vsp-production" \
                     --output-dir "./manifests/" \


chart/templates/gateway-ingress-gateway.yaml
```diff
@@ -27,9 +27,18 @@ spec:
...
       credentialName: {{ .Release.Name }}-gateway-ingress-certificate
-    hosts:
-    - {{ include "gateway.host" . }}
+{{ end -}}
+hosts:
+{{ if eq .Values.gateway.qwac.enabled -}}
+  - {{ include "gateway.host" . }}
+{{ else -}}
+  - {{ include "gateway.host.govuk" . }}
+{{ end -}}
```
idrop, line 44: I'd prefer one if/else block per case that contains the entire `tls` and `hosts` config, I'd find it easier to compare

chart/values.yaml
```diff
@@ -47,6 +47,8 @@ gateway:
     pullPolicy: IfNotPresent
   errorPageURL: https://www.integration.signin.service.gov.uk/proxy-node-error
   metadataPath: /ServiceMetadata
+  qwac:
+    enabled: false
```
idrop, line 51: could we move this up to `global`, as this switch may be used for connector node metadata hosting, which may not be in gateway.

Can we change the key to `qwacCertificate` to make it a bit clearer? 🤷‍♂️

JohnWattsGDS: I don't suppose it will take up much more of the RAM-Memory so it should be fine.  (Sorry, I'm a pedant when it comes to that sort of thing, I can change it to qwacCertificate)

chart/templates/gateway-ingress-gateway.yaml
idrop: I'd still prefer a bit of repetition and two distinct blocks i.e.
````
    {{- if eq .Values.global.qwacCertificate.enabled -}}
    tls:
      mode: SIMPLE
      credentialName: gateway-ingress-certificate-for-signin-sealed-secret
    hosts:
    - {{ include "gateway.host.govuk" . }}
    {{- else -}}
    tls:
      mode: SIMPLE
      serverCertificate: sds
      privateKey: sds
      credentialName: {{ .Release.Name }}-gateway-ingress-certificate
    hosts:
    - {{ include "gateway.host" . }}
    {{- end -}}
````
Also note the [chomp-left dash on the flow control statements](https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace)

Comment (JohnWattsGDS) [reply chart/templates/gateway-ingress-gateway.yaml]: That's what I did isn't it?  And I left the chomp left out as there's nothing to chomp but I'll add them in if you want.

Comment (JohnWattsGDS) [reply chart/templates/gateway-ingress-gateway.yaml]: I see, you want TLS inside too. Fair enough.

File: chart/templates/gateway-ingress-gateway.yaml [hunk 3]
```diff
@@ -26,10 +26,18 @@ spec:
...
+{{- else -}}
     tls:
       mode: SIMPLE
       serverCertificate: sds
       privateKey: sds
       credentialName: {{ .Release.Name }}-gateway-ingress-certificate
-    hosts:
-    - {{ include "gateway.host" . }}
+hosts:
+  - {{ include "gateway.host.govuk" . }}
```
idrop, line 42:
```suggestion
  - {{ include "gateway.host" . }}
```

chart/templates/gateway-ingress-gateway.yaml
```diff
@@ -26,10 +26,18 @@ spec:
       number: 443
       name: https
       protocol: HTTPS
+{{- if eq .Values.global.qwacCertificate.enabled -}}
```
blairboy362, line 29:
```suggestion
{{- if .Values.global.qwacCertificate.enabled -}}
```

blairboy362: Or:
```suggestion
{{- if eq .Values.global.qwacCertificate.enabled "true" -}}
```

chart/templates/gateway-ingress-gateway.yaml
```diff
@@ -26,10 +26,18 @@ spec:
...
+    tls:
+      mode: SIMPLE
```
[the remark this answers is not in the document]
JohnWattsGDS: 👍

PR #606, opened 2020-07-28
BAU Use parallelStream in MetadataResolverService constructor

JohnWattsGDS:
The MetadataResolverService needs to create a MetadataResolver for each configured country.  When it does this in series then any faulty countries must time out before the next country can be resolved.  As the resolvers are created during a stream operation we can replace `stream()` with `parallelStream()`.

Using the integration configuration, it can reduce the time taken to create the resolvers from ~42s to ~12s. Of the 12s, at least 10 are the time it takes for countries to timeout.  Reducing the timeout numbers in `getClient()` further speeds up the startup to ~8s.

You can test it by:

* generating a `countriesConfig.yaml` by running `verify-eidas-config.git/tools ./gradlew run --args="generate-config --environment=integration"`
* setting the environment variables `CONFIG_FILE` and `COUNTRIES_CONFIG_FILE` to suitable values (`src/dist/config.yml` for `CONFIG_FILE` and the file generated above for `COUNTRIES_CONFIG_FILE`)
* running the metatron `main` method.

**Example run configuration for IntelliJ**
image

### DIFF
--- a/metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
+++ b/metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
@@ -55,7 +55,7 @@ public class MetadataResolverService {
         this.metadataResolverFactory = metadataResolverFactory;
         this.credentialResolverFactory = credentialResolverFactory;
         this.expiredCertificateMetadataFilter = new ExpiredCertificateMetadataFilter();
-        this.countryConfigMap = countriesConfig.getCountries().stream()
+        this.countryConfigMap = countriesConfig.getCountries().parallelStream()
                 .collect(Collectors.toMap(EidasCountryConfig::getEntityId, this::createMetadataResolver));
     }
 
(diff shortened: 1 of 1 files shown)


metatron/src/main/java/uk/gov/ida/eidas/metatron/domain/MetadataResolverService.java
```diff
@@ -78,8 +78,8 @@ public CountryMetadataResponse getCountryMetadataResponse(URI entityId) {
 
     private Client getClient(EidasCountryConfig country) {
         ClientConfig configuration = new ClientConfig();
-        configuration.property(ClientProperties.CONNECT_TIMEOUT, 10000);
-        configuration.property(ClientProperties.READ_TIMEOUT, 10000);
+        configuration.property(ClientProperties.CONNECT_TIMEOUT, 5_000);
```
jkunle, line 81: Is this mean't to be 5_000 or 5000 is this milliseconds?

JohnWattsGDS: Yeah, 5_000 == 5000

It's a nice Java thing that you can add underscores to numbers like you would in writing like 5,000 and makes them a little more readable.

It is milliseconds also.

I briefly considered a constant like `TIMEOUT_MS = 5_000` but it seemed overkill.

PR #607, opened 2020-08-06
BAU upgrade CloudHSM client to latest version

idrop:
Follow instructions at https://docs.aws.amazon.com/cloudhsm/latest/userguide/install-and-configure-client-linux.html for Amazon Linux 2.

Cherry pick this onto the `release/single-country-legacy` branch once it passes the build.


instantiator:
The branch build passed, but PR build errored with `The command "npm install -g snyk@1.305.0" failed and exited with 1 during .` which seems odd and not particularly related to what you've changed - so perhaps it's actually fine and just a weird glitch?

PR #610, opened 2020-08-07
EID-1919 Second stab at a sealed secret

JohnWattsGDS:
Encrypt an actual certificate instead of a github page


blairboy362:
Someone with admin powers will also need to go in and delete the sealed secret and secret with the old name.

JohnWattsGDS:
I'll check with Phil that I haven't messed up the name too.

JohnWattsGDS:
I changed the name so does someone with admin powers still also need to go in and delete the sealed secret and "secret with the old name" (now that there isn't an old name).  Or rather - does anything need deleting or will it be overwritten?

blairboy362:
If the name is the same then it will just overwrite. No one will need to go in and tidy up afterwards.

PR #613, opened 2020-08-10
EID-1919 TIMTOWTDI

JohnWattsGDS:
But it's best to use just one way instead of two at a time.  Use a `generic` sealed secret with `--from-file=cert=...` and `--from-file=key=...` OR a `tls` sealed secret with `--from-file=tls.crt=...` and `--from-file=tls.key=...`.

:crossed_fingers:


Krenair:
we're basing this on https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/ and https://github.com/alphagov/gsp/blob/master/docs/gds-supported-platform/tls-certificates.md#using-a-custom-domain-by-providing-your-own-certificate

PR #621, opened 2020-08-14
BAU Test Metatron ingress/egress logging

JohnWattsGDS:
Add test for `metatron` ingress and egress logging including verifying logging of `PROXY_NODE_JOURNEY_ID` on ingress.

Modify AppRule to enable injection of `ProxyNodeJsonClient` when making a `GET` request (since `ProxyNodeJsonClient` passes the headers to `metatron` that need logging).

Modify `build.gradle` so that `ida_utils` dependencies are available to `testImplementation`.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -6,4 +6,8 @@ ignore:
     - '*':
         reason: Fix not available
         expires: 2020-12-15T00:00:00.000Z
+  SNYK-JAVA-ORGGLASSFISHJERSEYMEDIA-595972:
+    - '*':
+        reason: Not used directly by us; Fix not available
+        expires: 2020-09-07
 patch: {}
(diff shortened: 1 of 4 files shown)


.snyk
```diff
@@ -6,4 +6,8 @@ ignore:
     - '*':
         reason: Fix not available
         expires: 2020-12-15T00:00:00.000Z
+  SNYK-JAVA-ORGGLASSFISHJERSEYMEDIA-595972:
+    - '*':
+        reason: Not used directly by us; Fix not available
+        expires: 2020-09-07
```
j-maynard, line 12: How do we know if a fix will be available around this time?  Wouldn't say 3 months be a better length of time for this to expire?

JohnWattsGDS: Wishful thinking on my part / better to have to do it again in a month than not realise it's been fixed and have it unpatched for two months (assuming the fix becomes available).

j-maynard: Cool... I get that... just felt that it breaking again in a month was a bit sad if we could avoid it :-) :+1:

PR #623, opened 2020-08-18
EID-1994 Killswitch

JohnWattsGDS:
I don't really expect this to work but it might just have the parts it needs to do what we want it to.

Ideally we'd be able to see the lock file writing happen (lock/unlock) and then see if that happens when the killswitch is triggered.  Then we could add in a check for the lockfile to the deploy stuff.

### DIFF
--- a/ci/deploy/multi-country-killswitch-pipeline.yaml
+++ b/ci/deploy/multi-country-killswitch-pipeline.yaml
@@ -1,0 +1,207 @@
+---
+apiVersion: concourse.govsvc.uk/v1beta1
+kind: Pipeline
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: killswitch
+spec:
+  exposed: false
+  paused: true
+  config:
+
+    resource_types:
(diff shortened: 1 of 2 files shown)


ci/deploy/multi-country-killswitch-pipeline.yaml
```diff
@@ -0,0 +1,182 @@
...
+      - name: slack-notification
+        type: docker-image
+        source:
+          repository: cfcommunity/slack-notification-resource
+
+    resources:
+      - name: verify-slack
+        type: slack-notification
+        source:
+          url: ((slack-webhook))
```
idrop, line 28: On the GSP, I think this needs to be `((slack-notification.url))`, which is a sealed secret in this directory.
Also, I think that webhook slack url may be for #eidas-alerts

JohnWattsGDS: 👍

idrop (approved):

jkunle (approved):
LGTM

PR #628, opened 2020-08-21
uSe new govsvc/aws-ruby image

philandstuff:
This image is now continuously deployed, we should prefer it.  We're
moving away from the `gdsre` organisation to the `govsvc` one.


chrisfarms (approved):

idrop:
Thanks Phil. Merging as I want to PR on top of this.

PR #629, opened 2020-08-24
☠︎ EID-1994 use S3 resource in killswitch pipeline

idrop:
Use an `S3` concourse resource to reference a file (dis)allowing deployment to Proxy Node production.
This can be invoked via the `kill-and-lock-prod` or `lock-prod-deploy` jobs, and unlocked by other jobs.

The `S3` resource requires a namespaced Principal to hold access credentials (tested on sandbox).


j-maynard (approved):
:lion:

instantiator (approved):

PR #631, opened 2020-08-24
🛑 EID-1994 Stop Proxy Node Prod deploy if pipeline locked via S3

idrop:
Check the `lock-production` S3 bucket to see if someone has set a "lock" on the prod deploy pipeline.
Slack when the pipeline is locked, and the deploy won't happen.
Also remove the `force unlock` job which was a carry-over from the Hub pipeline as isn't used here.


JohnWattsGDS (approved):
🎉

instantiator (approved):

PR #639, opened 2020-09-09
BAU set prod error page url for Proxy Node

idrop:
Override the prod default error page [set in values.yaml](https://github.com/alphagov/verify-proxy-node/blob/0530d32c4327cfc1786b7b3543f0907d696a6c98/chart/values.yaml#L50). It's set to an integration url by default.

Also ignore a snyk vuln that is not used directly at runtime, for another month. We need to upgrade to Dropwizard 2 to mitigate.


JohnWattsGDS (approved):
🦖

AmritSidhu (approved):

PR #641, opened 2020-09-14
EID-2012 set eIDAS response issuer as Proxy Node entity id

idrop:
Set an `issuerId` field in `HubResponseTranslator`. Use this `issuerId` field to be the SAML `Issuer` of the response.
We're presently incorrectly setting the `Issuer` to be the country entity id.

The value of `issuerId` is helm templated into the translator deployment with `{{ include "gateway.entityID" . }}`

Unit, integration (apprule) and acceptance tests have been updated.

### DIFF
--- a/proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
+++ b/proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
@@ -88,7 +88,7 @@ And('they should have a transient PID') do
 end
 
 And('they should have a response issued by the Proxy Node') do
-  assert_text(env('PROXY_NODE_URL') + '/ServiceMetadata')
+  assert_text('Issuer: ' + env('PROXY_NODE_URL') + '/ServiceMetadata')
 end
 
 Then("the user should be presented with a Hub error page indicating IDP could not sign them in") do
--- a/proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
+++ b/proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
@@ -87,6 +87,10 @@ And('they should have a transient PID') do
   assert_text('GB/EU/_tr_')
 end
 
+And('they should have a response issued by the Proxy Node') do
+  assert_text(env('PROXY_NODE_URL') + '/ServiceMetadata')
+end
+
 Then("the user should be presented with a Hub error page indicating IDP could not sign them in") do
   assert_text('Stub Idp Demo One couldn’t sign you in')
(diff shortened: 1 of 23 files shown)


proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/HubResponseTranslatorRequest.java
```diff
@@ -39,7 +39,7 @@
 
     @NotNull
     @JsonProperty
-    private URI eidasIssuerEntityId;
+    private URI connectorEntityId;
```
Wynndow, line 42: Is this the `connectorEntityId` for the sending EU country? Maybe this could be something more specific to differentiate between their connector and ours. I know in this context it could only be theirs but it did make me think for a few seconds.

proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
```diff
@@ -87,6 +87,10 @@ def stub_idp_user
   assert_text('GB/EU/_tr_')
 end
 
+And('they should have a response issued by the Proxy Node') do
+  assert_text(env('PROXY_NODE_URL') + '/ServiceMetadata')
```
Wynndow, line 91: Maybe: `assert_text('Issuer: ' + env('PROXY_NODE_URL') + '/ServiceMetadata')` to be a bit sure it's appearing as expected?

Wynndow (dismissed):
Couple of comments, not deal breakers though. Feel free to ignore them!

jkunle (approved):

proxy-node-acceptance-tests/features/step_definitions/user_steps.rb
idrop: Good call, I'll change this

proxy-node-shared/src/main/java/uk/gov/ida/notification/contracts/HubResponseTranslatorRequest.java
idrop: My thought was this is the Proxy Node repo, so "connector" would always be the connecting country.

Wynndow: Yep, it's not really an issue. I can live live with it!

Wynndow (approved):

jkunle (approved):

PR #644, opened 2020-10-01
BAU Test for expired metadata entity descriptor

idrop:
Unit tests are failing as its past the expiry date on the test metadata template `connector_node_metadata_template.xml`.

Update this expiry date to the future, and introduce a new test that asserts expired metadata entity descriptor should make that metadata invalid.


Wynndow (approved):
🚀

JohnWattsGDS (approved):

PR #658, opened 2020-11-02
EID-2020: Sweden prod proxy node smoke

AmritSidhu:
Currently Sweden has an integration smoke test which use the same steps.
Named the first step 'SwedenProduction' as that is the easiest way to determine them apart.
It uses the same steps following that as the integration journey, it just uses a different url.

This feature can be tested locally by:

- updating the `run-tests.sh` on line 80 to point at `features/smoke/production`
- and then run the `run-tests.sh` script 

Subsequent PR in `verify-terraform`: https://github.com/alphagov/verify-terraform/pull/1340

https://govukverify.atlassian.net/browse/EID-2020


AmritSidhu:
There may be a nicer way to determine the different URL's, but this is the easiest/simplest way I initially thought of

Wynndow (approved):

idrop (approved):

PR #671, opened 2020-12-02
BAU: Fix Italian proxy node test

Wynndow:
The Italian proxy node test is _very_ flaky. We often see the test
failing at the Italians connector-node. There are a series of pages
which are auto-submitted with JS. One particular page's JS is fired by a
`load` event on the window. There may be a race condition as a lot of
the time this doesn't happen with Selenium and the journey stalls.

This adds a conditional check to see if we've stalled. If so, we
manually call the JS required to submit the page.


JohnWattsGDS (approved):

idrop (approved):
Thanks @Wynndow !

Wynndow:
@idrop No problem. I'm not going to admit how much time I've spend looking at this today....

PR #672, opened 2020-12-14
EID-2081 Proxy Node  integration kill switch

idrop:
## What
A way to remove the deployment of the Integration Proxy Node, via a concourse pipeline.

## Why
We already have a kill switch for production, but not integration. This will give us the confidence that the kill switch mechanics work, but won't affect production. We need more confidence as we may have to switch off the Proxy Node at a very specific out-of-hours time.

The change is mostly a copy-paste of the prod jobs, plus a new S3Bucket type to hold the integration lockfile.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,7 +5,7 @@ ignore:
   SNYK-JAVA-CAJULIUSDAVIES-30073:
     - '*':
         reason: Fix not available
-        expires: 2020-12-15T00:00:00.000Z
+        expires: 2021-01-01
   SNYK-JAVA-ORGGLASSFISHJERSEYMEDIA-595972:
     - '*':
         reason: Not used directly by us; Fix not available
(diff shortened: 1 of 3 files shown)


.snyk
```diff
@@ -5,7 +5,7 @@ ignore:
   SNYK-JAVA-CAJULIUSDAVIES-30073:
     - '*':
         reason: Fix not available
-        expires: 2020-12-15T00:00:00.000Z
+        expires: 2021-01-01
```
CharlesIC, line 8: Maybe not a good idea to make things expire on a holiday 😅

CharlesIC (approved):

.snyk
idrop: It should just affect travis checks on PRs, and we are thinking we won't have any post 2020.

CharlesIC: Oh fair

JohnWattsGDS (approved):

PR #678, opened 2021-01-19
BAU Remove Proxy Node build pipeline

idrop:
We don't need to build the Proxy Node anymore as we don't deploy it anywhere.


jkunle (approved):
Sad to see this go, all that time spent to craft this..

instantiator (approved):
👋

CharlesIC:
I imagine we'll archive the whole repo at some point - it might be worth keeping the Concourse pipelines for later reference

